PR #6, opened 2020-05-27
Musdoom: New Stats and Fix Bugs

JoanStinson:
What's new:

- Battlecry, hit and die behaviors + new animations integration
- Fixed issues with enemies behavior and animations


Scripts/src/Script/EnemyState.cpp
```diff
@@ -18,7 +20,9 @@ EnemyState::EnemyState(Mushdoom* enemy) : enemy(enemy)
 
 void EnemyState::Exit(EnemyState* state)
 {
+	enemy->animation->Stop();
 	enemy->current_state->OnStateExit();
 	enemy->current_state = state;
 	enemy->current_state->OnStateEnter();
+	enemy->animation->Play();
```
Dartemiss, line 27: We should take a look at this in the future

JoanStinson: Yeah, not sure, but this solves the freeze issue :v

PR #7, opened 2020-05-27
Feature/light ball

JorxSS:
First iteration with Light Ball.
You will find it in the MainSceneWithHUD.

- In order to use it you will need the last update from develop in the Engine.

The Light Ball is spawned with Keyboard I or Gamepad clicking Right Joystick.
Also the light ball do some damage to the enemies.


Scripts/src/Script/LightBall.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+}
+
+bool LightBall::UpdateProjectile()
+{
+	float3 ball_distance = force_position + float3::unitY - owner->transform.GetGlobalTranslation();
+	expected_velocity = ball_distance.Normalized() * speed;
+	steering = expected_velocity - current_velocity;
+
+	current_velocity += steering;
```
mariofv, line 184: This is the same as `current_velocity = eexpected_velocity`

Scripts/src/Script/LightBall.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+	current_velocity *= slow_down_factor;
+
+	next_position = owner->transform.GetGlobalTranslation() + current_velocity * App->time->delta_time;
+	owner->transform.SetGlobalMatrixTranslation(next_position);
+
+	DetectCollision();
+
+	return ball_distance_magnitude < range;
+}
```
mariofv, line 218: It's better to have a single method with the movement logic, and just change the target where you are going

Scripts/src/Script/PlayerAttack.cpp
```diff
@@ -43,9 +43,9 @@ void PlayerAttack::Awake()
 	collider_component = static_cast<ComponentCollider*>(collider->GetComponent(ComponentCollider::ColliderType::BOX));
 	audio_source = static_cast<ComponentAudioSource*>(owner->GetComponent(Component::ComponentType::AUDIO_SOURCE));
 
-	GameObject* ui = App->scene->GetGameObjectByName("UIManager");
-	ComponentScript* component_ui = ui->GetComponentScript("UIManager");
-	ui_manager = static_cast<UIManager*>(component_ui->script);
+	//GameObject* ui = App->scene->GetGameObjectByName("UIManager");
+	//ComponentScript* component_ui = ui->GetComponentScript("UIManager");
+	//ui_manager = static_cast<UIManager*>(component_ui->script);
```
mariofv, line 48: If it's not used anymore better delete it to keep the code cleaned

Issue #3, opened 2020-05-27
[BUG] Camera freezes at weird position in Singleplayer


JoanStinson:
**Describe the bug**
If when entering a trigger (that freezes the camera and spawn enemies) while jumping or moving very far away of the Z axis, touching the border, when the camera freezes, it freezes at a weird position.

**To Reproduce**
1. Singleplayer
2. When entering a trigger jump or move at the maximum Z the path allows you.

**Expected behavior**
The camera should freeze at the same position always, regardless if you jump or are far away of the camera.

**Screenshots**
Trigger 1: https://gyazo.com/f81e70ce9576ba0eb648914d102e1e9a
Trigger 2: https://gyazo.com/7c47d504b6bf53e1b2035f8a795d9a36

PR #8, opened 2020-05-27
Bug/fix camera bug #3

JorxSS:
fixed camera bug reported.

Now there is an specific offfset for the battle triggers and transitions.


Scripts/src/Script/CameraController.cpp
```diff
@@ -246,5 +250,14 @@ void CameraController::InitPublicGameObjects()
 
 void CameraController::SetPosition(float3 position)
```
JoanStinson, line 251: To avoid copying values preferably pass a reference. In this case you are not modifying the value, so it should be a const reference! 😃 Like: `void CameraController::SetPosition(const float3& position)`.

Scripts/src/Script/LightBall.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+LightBall::LightBall()
+{
+	panel = new PanelComponent();
+}
+
+// Use this for initialization before Start()
+void LightBall::Awake()
+{
+	GameObject* world_go = App->scene->GetGameObjectByName("World Manager");
```
JoanStinson, line 38: Why the space here `("World Manager")`? And on the others not? 👀 haha

Scripts/src/Script/LightBall.cpp
```diff
@@ -0,0 +1,258 @@
...
+// Use this for initialization
+void LightBall::Start()
+{
+	if (world_manager->singleplayer)
+	{
+		if (!world_manager->player1_choice)
+		{
+			current_player = player1;
+			next_target = player1;
+			current_player_controller = player1_controller;
```
JoanStinson, line 67: If you are assigning 2 values to the same one you can make it with 1 line of code instead of 2. 
Instead of doing this:
```
current_player_controller = player1_controller;
next_target_controller = player1_controller;
```
**Do this:** 
`current_player_controller = next_target_controller = player1_controller;`

Scripts/src/Script/LightBall.cpp
```diff
@@ -0,0 +1,258 @@
...
+			next_target_controller = player1_controller;
+		}
+		else
+		{
+			current_player = player2;
+			next_target = player2;
+			current_player_controller = player2_controller;
```
JoanStinson, line 74: The same as ☝️

Scripts/src/Script/LightBall.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+bool LightBall::UpdateProjectile()
+{
+	float3 ball_distance = force_position + float3::unitY - owner->transform.GetGlobalTranslation();
+	expected_velocity = ball_distance.Normalized() * speed;
+	steering = expected_velocity - current_velocity;
+
+	current_velocity += steering;
+
+	ball_distance_magnitude = sqrt(pow(ball_distance.x, 2) + pow(ball_distance.y, 2) + pow(ball_distance.z, 2));
```
JoanStinson, line 186: I think it was Carlos with the vector class exercise that he said that on the CPU it's better to do the pow manually if it's not too high because it's more optimal. So, instead of doing `pow(ball_distance.x, 2)` directly write `(ball_distance.x * ball_distance.x)` (for the y and z also)

Scripts/src/Script/LightBall.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+bool LightBall::UpdateTransition()
+{
+	float3 ball_distance = next_target->transform.GetGlobalTranslation() + float3::unitY - owner->transform.GetGlobalTranslation();
+	expected_velocity = ball_distance.Normalized() * speed;
+	steering = expected_velocity - current_velocity;
+
+	current_velocity += steering;
+
+	ball_distance_magnitude = sqrt(pow(ball_distance.x,2) + pow(ball_distance.y, 2) + pow(ball_distance.z, 2));
```
JoanStinson, line 207: The same as ☝️

Scripts/src/Script/WorldManager.cpp
```diff
@@ -274,6 +274,7 @@ void WorldManager::CheckTriggers()
 			if(!event_manager->events_triggered[i])
 			{
 				event_manager->TriggerEvent(i);
+				camera_manager->SetPosition(event_triggers[i]->owner->transform.GetGlobalTranslation() + float3(0.f, 8.5f, 20.5f));
```
JoanStinson, line 277: `float3(0.f, 8.5f, 20.5f)` magic number? Variable better :))

Scripts/src/Script/LightBall.cpp
JorxSS: The GO is called World Manager in the Hierarchy

Scripts/src/Script/LightBall.cpp
JorxSS: this was from another PR already modified.

Scripts/src/Script/LightBall.cpp
JorxSS: this was from another PR already modified.

Scripts/src/Script/LightBall.cpp
JorxSS: this was from another PR already modified.

Scripts/src/Script/LightBall.cpp
JorxSS: this was from another PR already modified.

PR #9, opened 2020-05-29
Feature/movement

marcorod94:
- Now the movement game objects is done by the transform of it.
- Collision detection is still working with bullet.


mariofv:
Closing this because we have a new approach

PR #10, opened 2020-06-01
Feature/refactor debug

JorxSS:
When Debug Mode enable:
* Teleport to pre-set zones. (In order to do that, you need to make a DebugSystem's empty game object child, and place it wherever you want)
* Invincible mode now have a damage bonification against enemies.
* Spawn Mushdooms in front of you.

Known Bugs:
* If you TP to the platform zone more than 1 time you will not trigger the platform camera
* If you are singleplayer but select the Female Character the debug mode is not going to work as expected.

This known bugs will be solved with the current Manager's refactor


Scripts/src/Script/PlayerAttack.cpp
```diff
@@ -100,7 +105,14 @@ void PlayerAttack::ComputeCollisions() const
...
 			}
-			enemy->TakeDamage(current_damage_power);
+			if(debug->is_player_invincible)
+			{
+				enemy->TakeDamage(current_damage_power * 100);
+			}
+			else
+			{
+				enemy->TakeDamage(current_damage_power);
+			}
```
Dartemiss, line 115: Uff that's nasty, I would just change the value of current_damage_power when invincible for code clarity.

PR #15, opened 2020-06-03
Feature/revive

JorxSS:
Now you can revive a dead character.
 - Until game input modified for both players you need to click V. (Will be changed with the game input.)
- Make a player die, and with the other player go for it and click V during 3 seconds.


Scripts/src/Script/PlayerController.cpp
```diff
@@ -45,44 +47,104 @@ void PlayerController::Awake()
...
+		else
+		{
+			//TODO Change the input with Revive Game Input
+			if(abs(current_distance) <= revive_range && App->input->GetKeyDown(KeyCode::V))
+			{
+				current_state = PlayerState::REVIVING;
+				current_time = App->time->delta_time;
+				revive_bar->SetProgress(current_time / revive_time);
+				revive_bar->owner->SetEnabled(true);
+			}
```
Dartemiss, line 125: I don't understand this, the player who is dead is the one who looks for the input? Shouldn't be the opposite? Because this way the dead player could revive himself by pressing its key.

Scripts/src/Script/PlayerController.cpp
```diff
@@ -172,6 +238,56 @@ void PlayerController::Die()
...
+	current_state = PlayerState::DEAD;
+	owner->children[3]->SetEnabled(true);
+	
+	//owner->SetEnabled(false);
+}
+
+float PlayerController::CheckDistance()
+{ 
+	float3 distance = other_player->player_go->transform.GetGlobalTranslation() - owner->transform.GetGlobalTranslation();
+	return distance.x;
```
Dartemiss, line 252: Why only checking x distance?

Scripts/src/Script/PlayerController.cpp
```diff
@@ -172,6 +238,56 @@ void PlayerController::Die()
...
+}
+
+void PlayerController::FollowOtherPlayer()
+{
+	float3 distance = other_player->player_go->transform.GetGlobalTranslation() - owner->transform.GetGlobalTranslation();
```
Dartemiss, line 257: This should be direction more than distance (just naming thing)

Scripts/src/Script/PlayerController.cpp
```diff
@@ -172,6 +238,56 @@ void PlayerController::Die()
...
 
+void PlayerController::Revive()
+{
+	//Here make all related to death after die animation
+	is_alive = true;
+	owner->children[0]->SetEnabled(true);
+	collider->active_physics = true;
+	current_state = PlayerState::ALIVE;
+	owner->children[3]->SetEnabled(false);
+	health_points = 500.f;
```
Dartemiss, line 280: I think that this shouldn't be harcoded, I believe it was max_health / 4.f

Scripts/src/Script/PlayerController.cpp
Dartemiss, line 252: This function should be const :D

Scripts/src/Script/PlayerController.cpp
JorxSS: the input is going to take the other players id to check his input (but game input is not made yet), I thought this way will be easier than everytime check if I'm near a player and check if it's alive and then put all the logic.

Scripts/src/Script/PlayerController.cpp
JorxSS: from the point of view of the player is the easiest point. Was my thought.

Scripts/src/Script/PlayerController.cpp
JorxSS: Done

Scripts/src/Script/PlayerController.cpp
JorxSS: True! Done

Scripts/src/Script/PlayerController.cpp
JorxSS: I put that value because at that moment I didn't knew if would be half or quarter, also I read that the player alive is going to lose some health so that's a TODO.

JorxSS: Done

PR #16, opened 2020-06-05
manual movement

vidalmelero:
full player movement with transform. Manual collision with mesh collider.


mariofv:
Closing this because we have a new approach

PR #19, opened 2020-06-10
Feature/input refactor

Dartemiss:
What's new?
- New input manager script that is the one who manages all the inputs and calls the module input of the engine. This way the module input only cares about handling the inputs from sdl and is this script the one that takes care of number of controllers/players/multiplayer etc
- Use triggers as buttons
- Copied the models and scenes from combo branch in order to not crash


Scripts/src/Script/InputManager.h
```diff
@@ -0,0 +1,41 @@
...
+	float GetHorizontalRaw(PlayerID player_id) const;
+
+	bool GetGameInput(const char* name, PlayerID player_id);
+	bool GetGameInputDown(const char* name, PlayerID player_id);
+	bool GetGameInputUp(const char* name, PlayerID player_id);
+
+	bool singleplayer_input = true;
+	int total_game_controllers = 0;
+
+	WorldManager* world_manager = nullptr;
```
mariofv, line 37: Variables are not separated fro methods

PR #25, opened 2020-06-13
Feature/combos

JorxSS:
[HIGHLIGHTS]
- New Queue Inputs.
- Amazing Attacks.
- New combos.
- Dash.

TODO: Add game inputs


Scripts/src/Script/PlayerAttack.cpp
```diff
@@ -55,38 +57,82 @@ void PlayerAttack::Awake()
...
 {
-	is_attacking = animation->IsOnState("Punch") || animation->IsOnState("Kick");
 
-	if(!is_attacking)
+	anim_percentatge = animation->GetCurrentClipPercentatge();
+
+	if(on_cooldown)
 	{
-		if (input_manager->GetGameInputDown("Punch", static_cast<PlayerID>(player - 1)))
+		current_combo_cooldown += App->time->delta_time / 1000.f;
```
marcorod94, line 76: it would be better to multiply for 0.001F and maybe have a global variable for it instead of use magic numbers

Scripts/src/Script/PlayerAttack.cpp
```diff
@@ -152,3 +231,134 @@ void PlayerAttack::InitPublicGameObjects()
...
+	if(current_attack->id == 0)
+	{
+		//Check if Button matches with some transition and modify current attack (strange)
+		if(UpdateAttack(input->button))
+		{
+			animation->ActiveAnimation(current_attack->animation);
+			input_queue.clear();
+			attack_activated = true;
+		}
+	}
```
marcorod94, line 329: Could this block be on if instead of two?

Scripts/src/Script/PlayerController.cpp
```diff
@@ -105,23 +105,49 @@ void PlayerController::Update()
...
+					is_dashing = false;
+					current_dash_cooldown = dash_cooldown;
+				}
+
+				return;
 			}
+			if(!is_dashing)
+			{
+				if (!is_attacking)
+				{
```
marcorod94, line 126: This could be the else part instead of a different if

Scripts/src/Script/PlayerAttack.cpp
Dartemiss: Done

Scripts/src/Script/PlayerAttack.cpp
Dartemiss: No because we need to control the else for the case when it is already on combo

Scripts/src/Script/PlayerController.cpp
Dartemiss: The else part of what? They are different booleans is_attacking and is_dashing

PR #63, opened 2020-07-02
Improved attack collision

Dartemiss:
What's new?
- Now we don't use collider for attacking, we use pure math!
- Implemented @JorxSS approach for combos with placeholders.


Scripts/src/Script/PlayerAttack.cpp
```diff
@@ -134,30 +133,15 @@ bool PlayerAttack::Attack(int player)
...
-			original_position = true;
-		}
-
 		if(!current_attack->is_aerial && current_attack->velocity > 0.f)
 		{
 			player_movement->MoveIntoDirection(owner->transform.GetFrontVector(), current_attack->velocity);
 		}
 		else if(current_attack->is_aerial && (current_attack->velocity || 0.f) && !player_movement->IsGrounded()
-			&& animation->IsOnState("DownTailSmash") && animation->GetCurrentClipPercentatge() > 0.3f)
+			&& animation->IsOnState("DownTailSmash") && anim_percentatge > 0.3f)
```
JoanStinson, line 143: `0.3f` magic number ✨

PR #64, opened 2020-07-04
Feature/camera shake

Dartemiss:
-Whats new?
- Tunable camera shake using perlin noise.
- Translation and rotation shake available.

PD: USE DEVELOP


Scripts/src/Script/EnemyController.cpp
```diff
@@ -316,3 +316,13 @@ void EnemyController::SetMarkEnabled(bool enabled)
 {
 	warp_mark->SetEnabled(enabled);
 }
+
+float EnemyController::GetCurrentHealth() const
+{
+	return health;
+}
+
+float EnemyController::GettTotalHealth() const
```
Pacasasgar, line 325: There is a 'typo' here with the double 't' in 'Gett'.

Pacasasgar:
Camera shake works fine, gj!
Before approving there are merge conflicts with the 'level1.scene'.

JorxSS:
I think the Trauma should be modifiable from inspector, but GJ.
image

PR #68, opened 2020-07-06
Feature/fix rotation player

JorxSS:
WHAT'S NEW?

* Now the player rotates always even when he cannot move.
*  Player is not longer higher than the ground. Is always touching in it. (Check the shadow)

KNOWN BUGS:
First rock. The collider have a little ramp so it scales it. How to solve it? That zone needs to be totally perpendicular.


Scripts/src/Script/PlayerMovement.cpp
```diff
@@ -315,10 +315,7 @@ void PlayerMovement::ComputeVerticalVelocity(float3 direction, float speed, bool
 	bool falling = velocity.y < 0.f;
 	//grounded = collider->IsGrounded();
 
-	float3 origin = collider->GetOrigin();
-	float3 end = origin;
-	end.y -= collider->box_size.getY();
-	ComponentCollider* ground_check = collider->RaycastClosestHit(origin, end).collider;
+	ComponentCollider* ground_check = collider->DetectCollisionWithGround(0.75f).collider;
```
JoanStinson, line 318: `0.75f` Magic number? ✨

marcorod94: Same

PR #71, opened 2020-07-07
Feature/enemies knockback

Dartemiss:
-What's new?
- Mushdoom knockback implemented (biter will be done when @Pacasasgar & @JordiRomagosa finish it is easy).
- Distance of the knockback is computed depending on the damage the player is dealing and the speed of the movement depends on the time of hit animation of the mushdoom.


Scripts/src/Script/Mushdoom.cpp
```diff
@@ -204,6 +205,7 @@ void Mushdoom::TakeDamage(float damage)
 	}
 	else
 	{
+		knockup_power = damage / 10.f;
```
JoanStinson, line 208: Why divided by 10? To make it between 0-1? Btw `10.f` magic number ✨

Scripts/src/Script/MushdoomStateHit.h
```diff
@@ -16,7 +16,9 @@ class MushdoomStateHit : public EnemyState<Mushdoom>
 	void OnStateExit() override;
 
 private:
-	float total_time_hitted = 0.7f;
+	float total_percentatge_hitted = 0.95f;
```
JoanStinson, line 19: Correctly you spell it `percentage`, but this has nothing to do with code, so don't worry haha

Scripts/src/Script/Mushdoom.cpp
```diff
@@ -405,4 +412,38 @@ void Mushdoom::CancelAttack()
...
+	
+	float speed = init_move_speed;
+	float3 final_velocity = new_position;
+
+	bool falling = new_position.y < 0.f;
+	float3 origin = collider->GetOrigin();
+	float3 end = origin;
+	end.y -= collider->box_size.getY();
+	ComponentCollider* ground_check = collider->RaycastClosestHit(origin, end).collider;
+	bool grounded = ground_check != nullptr && ground_check->owner->tag == "EnvironmentCollider";
```
marcorod94, line 430: Use the specific function for detecting collision with ground

PR #73, opened 2020-07-09
Feature/custom area of damage

Dartemiss:
-What's new?
- Now attacks have their own values of attack_range and angle.

This stills need balancing.


Scripts/src/Script/DebugModeScript.cpp
```diff
@@ -197,6 +198,12 @@ void DebugModeScript::DrawDebugFeatures()
 		ImGui::Checkbox("Draw Global bounding boxes", &show_global_bounding_boxes);
 		ImGui::Checkbox("Draw Physics", &show_physics);
 		ImGui::Checkbox("Draw Movement Indicators", &show_movement);
+		ImGui::Checkbox("Draw Player1 Combat Indicators", &player_1->player_controller->player_attack->debug_collision_draw);
+		
+		if(player_2->player_controller->player_attack)
+		{
+			ImGui::Checkbox("Draw Player2 Combat Indicators", &player_2->player_controller->player_attack->debug_collision_draw);
+		}
```
JorxSS, line 206: you could check here if it's multiplayer instead

Dartemiss: Well this way I ensure that also it is not nullptr

PR #74, opened 2020-07-09
Feature/ui dialogues

JorxSS:
What's new:

* Dialogue System. 
* Camera improvements for multiplayer.

HOW TO TEST:
Press U in order to start the dialogues. Then pass it with the SpaceBar or with the lower button (A in Xbox Controller, and X in PlayStation Controller) and then see the results.

It's just the system, I left it for the UI and artist to make it beauty.


Scripts/src/Script/CameraController.cpp
```diff
@@ -62,11 +70,16 @@ void CameraController::Update()
 		return;
 	}
 
-	level_path->DrawLine();
+	if(debug_options)
+	{
+		level_path->DrawLine();
+		camera_focus->DrawLine();
+		camera_rail->DrawLine();
+	}
```
Dartemiss, line 78: Why is this on this pr?

Scripts/src/Script/DialogueManager.cpp
```diff
@@ -0,0 +1,191 @@
...
+				if(spawn_time < 0.f)
+				{
+					UpdateText();
+					spawn_time = cooldown_spawn;
+				}
+				
+			}
+		}
+	}
+	else if(App->input->GetKeyDown(KeyCode::U))
```
Dartemiss, line 94: This is hardcoded for testing I believe

Scripts/src/Script/CameraController.cpp
JorxSS: because I touched also the camera, is in the description mi pana

Scripts/src/Script/DialogueManager.cpp
JorxSS: yes, this is for testingu purposes

PR #82, opened 2020-07-12
Bugfix/dash and jump

marcorod94:
What's new
- Fix jump and dash issue
- Remove second jump
- Remove some unused variables from player movement


Scripts/src/Script/PlayerController.cpp
```diff
@@ -119,6 +119,15 @@ void PlayerController::Update()
 		case PlayerState::ALIVE:
 			if(!is_reviving)
 			{
+				if (CheckDashStatus() && ((App->input->GetKeyDown(KeyCode::LeftShift) && player == 1) ||
+					(input_manager->GetGameInputDown("Dash", static_cast<PlayerID>(player - 1)))))
```
Dartemiss, line 123: Why is repeated the input? I suppose GetGameInputDown("Dash") already contains leftshift

marcorod94: Ready

PR #85, opened 2020-07-13
Biter behaviours

Pacasasgar:
Biter logic added **without animations**:
-It chases the player.
-It can 'bite' as a normal attack (this attack can affect both player in a low radius).
-It can charge through one player's position and damage all players staying at its path (this attack has cooldown). During its duration the Biter remains invincible.
-It can make an explosion around itself after a short-cast damaging all players within its range (this attack has cooldown).
-It can receive damage and die.

Co-author ->@JordiRomagosa


Scripts/src/Script/EnemyController.cpp
```diff
@@ -180,6 +213,25 @@ void EnemyController::Seek(const float3& velocity) const
...
+	Quat new_rotation = Quat::Slerp(owner->transform.GetRotation(), desired_rotation, App->time->delta_time * 0.01F);
+	owner->transform.SetRotation(new_rotation);
+
+	float3 offset = float3(0, 1 + math::Sin(App->time->time / 300) / 5 + math::Sin(App->time->time / 150) / 20, 0);
+	float3 movement = current_velocity;
+	movement = movement.Cross(float3(0, 1, 0));
+	offset += movement * math::Sin(App->time->time / 350) / 3 + movement * math::Sin(App->time->time / 150) / 20;
+	owner->transform.SetGlobalMatrixTranslation(current_position + offset);
+}
+
```
JorxSS, line 234: this should be moved to Bitter.cpp

Scripts/src/Script/EnemyController.cpp
```diff
@@ -233,12 +294,54 @@ bool EnemyController::PlayerInAttackRange() const
 	return target.Distance(owner->transform.GetGlobalTranslation()) <= attack_range;
 }
 
+bool EnemyController::PlayerInAttackArea(const Sphere &sphere) const
+{
+	float3 target = current_target->owner->transform.GetGlobalTranslation();
+	return target.Distance(sphere) == 0;
+}
+
+bool EnemyController::PlayerInAttackArea(const Capsule & capsule) const
```
JorxSS, line 303: code style here `const Capsule& capsule`

Scripts/src/Script/EnemyController.cpp
```diff
@@ -233,12 +294,54 @@ bool EnemyController::PlayerInAttackRange() const
...
 	return target.Distance(owner->transform.GetGlobalTranslation()) <= range;
 }
 
+void EnemyController::DealDamageToPlayersInArea(const Sphere & sphere, float damage) const
+{
+	bool p1 = false; bool p2 = false;
+	DealDamageToPlayersInArea(sphere, damage, p1, p2);
+}
+
+void EnemyController::DealDamageToPlayersInArea(const Sphere & sphere, float damage, bool &player_1_hit, bool &player_2_hit) const
```
JorxSS, line 321: same here, code style typevariable& variable, not viceversa

Scripts/src/Script/EnemyController.cpp
```diff
@@ -233,12 +294,54 @@ bool EnemyController::PlayerInAttackRange() const
 	return target.Distance(owner->transform.GetGlobalTranslation()) <= attack_range;
 }
 
+bool EnemyController::PlayerInAttackArea(const Sphere &sphere) const
```
JorxSS, line 297: cs here, and why do you need two methods called the same, if at the end the attack collision is calculated with pure math.?
Will not be easier to make if it's just in front, do a dot product with the front vector of the enemy, and if it's an sphere just a distance vector with threshold?

Scripts/src/Script/EnemyController.h
```diff
@@ -40,17 +41,23 @@ class EnemyController : public Script
...
 	void MoveTowardsPlayer() const;
 	float3 CollisionAvoidance(const float3& velocity) const;
 
 	bool PlayerInSight() const;
 	bool PlayerInAttackRange() const;
+	bool PlayerInAttackArea(const Sphere &sphere) const;
+	bool PlayerInAttackArea(const Capsule &capsule) const;
 	bool PlayerInRange(float range) const;
+	void DealDamageToPlayersInArea(const Sphere &sphere, float damage) const;
+	void DealDamageToPlayersInArea(const Sphere &sphere, float damage, bool &player_1_hit, bool &player_2_hit) const;
```
JorxSS, line 56: again why you have 2 functions called equally and one just calls the other init 2 variables and calls the other one?

JorxSS, line 56: code style here

Scripts/src/Script/EnemyController.h
```diff
@@ -13,6 +13,7 @@ class ComponentParticleSystem;
 class EnemyManager;
 class LightOrbsManager;
 class PlayerController;
+class DebugModeScript;
```
JorxSS, line 16: alphabetical order

Scripts/src/Script/EnemyController.cpp
```diff
@@ -16,6 +16,34 @@
...
 
 
+#include "imgui.h"
+
+#include "EnemyManager.h"
+#include "LightOrbsManager.h"
+#include "PlayerController.h"
+#include "UIManager.h"
+#include "WorldManager.h"
+#include "DebugModeScript.h"
```
JorxSS, line 26: alphabetical order here

Scripts/src/Script/EnemyController.cpp
JordiRomagosa: It is not the same to do it how you say. It could work for the attacks where the enemy is at the center, but any other attack would not. For example: for the bite that is in front of the enemy. Or imagine any other attack an enemy can have that has an area of effect.
This is a function to see if the player is inside the area, maybe it can be refactored and is not needed (only use the deal damage in area function all time), but we are working on the premise of "having something working as soon as possible" and the refactor will be done later anyway.

Scripts/src/Script/EnemyController.h
JordiRomagosa: Because they have different uses. The first only deals damage, the second returns if one of the players has been hit.
I wanted to create a single one with a default value for the bool, but couldnt manage to do so since they are references. And the bool references for the second one are needed for attacks that have a duration and need to check the damage multiple times, so that the damage is controlled on the enemy attack that calls this. This way the enemy can know how many times a player is hit (counting each time this returns true), which one is hit, or if one of them is hit at all or not. And also if the player can take damage with the attack, with the value of the reference you send at the start.

Scripts/src/Script/EnemyController.cpp
JorxSS: but I checked the code and the function with the capsule is not used anywhere, so should be removed. And then you could even remove the creation of the Sphere, and just do `owner->transform.GetGlobalTranslation().Distance(current_target->transform.GetGlobalTranslation())` and that would be it, without the need to create an sphere object.

Scripts/src/Script/EnemyController.h
JorxSS: but the same way you put two variables just to false, call that funcion that doesn't need the booleans to false and false and that would be it. 
Also the same with the sphere, as I commented before, you could avoid the creation of the sphere and just get distnce from owner and each player, less complexity.

Scripts/src/Script/BiterStateAttack.cpp
```diff
@@ -0,0 +1,59 @@
...
+	enemy->is_attacking = true;
+	attacked = false;
+}
+
+void BiterStateAttack::OnStateUpdate()
+{
+	float current_percentage = enemy->animation->GetCurrentClipPercentatge();
+	float3 attack_center = enemy->current_position + (enemy->current_velocity.Normalized() * enemy->bite_distance);
+
+	if (current_percentage >= 0.40f && current_percentage <= 0.60f && !attacked)
```
JoanStinson, line 33: Good practice to store this numbers `0.40f` and `0.60f` as private members of the class to avoid magic numbers! ✨

Scripts/src/Script/BiterStateAttack.cpp
```diff
@@ -0,0 +1,59 @@
...
+	{
+		enemy->App->debug_draw->RenderSphere(attack_center, enemy->bite_radius);
+	}
+}
+
+void BiterStateAttack::OnStateExit()
+{
+	enemy->is_attacking = false;
+	//enemy->CancelAttack();
+	//enemy->LeaveAttackRange();
```
JoanStinson, line 58: Erase comments

Scripts/src/Script/BiterStateCharge.cpp
```diff
@@ -0,0 +1,54 @@
...
+void BiterStateCharge::OnStateEnter()
+{
+	if (!enemy->animation->IsOnState("Charge"))
+	{
+		enemy->animation->ActiveAnimation("charge");
+		//audio_source->PlayEvent("play_charge_biter");
+	}
+
+	p1_attacked = false;
+	p2_attacked = false;
```
JoanStinson, line 25: You can put it directly in one line of code like this: `p1_attacked = p2_attacked = false;`

Scripts/src/Script/BiterStatePursue.cpp
```diff
@@ -0,0 +1,51 @@
...
+		Exit(enemy->explosion_state);
+	}
+	/*
+	// Animation has 60 frames, steps each 15 frames (1, 15, 30, 45)
+	float current_percentage = enemy->animation->GetCurrentClipPercentatge();
+
+	if ((int)(current_percentage * 100) % 25 == 0 && current_percentage != 1)
+	{
+		enemy->audio_source->PlayEvent("play_footstep");
+	}*/
```
JoanStinson, line 46: Erase comments

Scripts/src/Script/EnemyController.cpp
```diff
@@ -95,6 +123,11 @@ void EnemyController::InitMembers()
 	world_manager = static_cast<WorldManager*>(App->scene->GetGameObjectByName("World Manager")->GetComponentScript("WorldManager")->script);
 
 	obstacles = App->scene->GetGameObjectsWithTag("Obstacle");
+	env_colliders = App->scene->GetGameObjectsWithTag("EnvironmentCollider");
+
+	GameObject* debug_system = App->scene->GetGameObjectByName("DebugSystem");
+	const ComponentScript* component_debug = debug_system->GetComponentScript("DebugModeScript");
+	debug = (DebugModeScript*)component_debug->script;
```
JoanStinson, line 130: Directly put it in one line of code like this: `debug = (DebugModeScript*)App->scene->GetGameObjectByName("DebugSystem")->GetComponentScript("DebugModeScript")->script;`

Scripts/src/Script/EnemyController.cpp
```diff
@@ -95,6 +123,11 @@ void EnemyController::InitMembers()
...
 
 	//Get Components
 	for (const auto go : owner->children)
```
JoanStinson, line 133: When using this type of for loop always use references '&' to avoid copies. So change to `for (const auto& go : owner->children)`

Scripts/src/Script/EnemyController.h
JoanStinson: Put this functions in the Biter.cpp!

Pacasasgar:
Now we have 5 animations and the state machine is working fine. We have also improved some parts and Jordi added particles to the Biter. These particles helps the player to identify better the attacks form the biter and 'react' to them. It's easier now to understand when the Biter is charging/exploding/basic_attacking with the animations and particles.

Scripts/src/Script/Biter.cpp
```diff
@@ -1,88 +1,374 @@
...
+		if (GetClosestTarget() == nullptr && current_state != idle_state)
+		{
+			current_state->Exit(idle_state);
+		}
+
+		if (activate_timer)
+		{
+			current_time += App->time->delta_time;
+
+			if (current_time > (seconds_to_disappear/* * 1000*/))
```
JorxSS, line 121: why is commented  `/* * 1000*/` ?

Scripts/src/Script/Biter.cpp
```diff
@@ -1,88 +1,374 @@
...
+}
+
+void Biter::TakeDamage(float damage)
+{
+	if (!is_alive || animation->IsOnState("Die") || animation->IsOnState("Hit")) return;
+
+	if (is_charging)
+	{
+		ui_manager->SpawnDamageIndicator(0, owner->transform.GetGlobalTranslation());
+	}
```
JorxSS, line 173: if it's charging a Damage indicator it's spawning but doesn't receive damage?

Scripts/src/Script/Biter.cpp
```diff
@@ -220,8 +220,9 @@ void Biter::MoveIntoDirection(float3 direction, float speed)
 	//ComputeVerticalVelocity(float3(0.f, normalized_direction.y, 0.f), speed, false);
 
 	//ApplyVelocity();
-	current_position = owner->transform.GetGlobalTranslation() + velocity * App->time->delta_time * 0.001f;
+	current_position = current_position + velocity * App->time->delta_time * 0.001f;
```
JorxSS, line 279: `current_position += velocity * App->time->delta_time * 0.001f;`

Scripts/src/Script/Biter.cpp
Pacasasgar: Because the Biter is supposed to disappear on flames with VFX when he dies, but right now we don't need to wait for the effects for the Biter to disappear after 'death' animation finishes because there is no VFX for that yet

Scripts/src/Script/Biter.cpp
Pacasasgar: During the charge the Biter should receive no damage. This is to indicate that you dealt 0 damage to him during the charge.

Scripts/src/Script/Biter.cpp
JorxSS: okay, and I assume that seconds_to_disappear  it's used in another place, if that is not the case you could just make seconds_to_disappear bigger

Scripts/src/Script/Biter.cpp
JorxSS: okay perfect then!!

Scripts/src/Script/Biter.cpp
Pacasasgar: Yes, it's also used for the Mushdoom (actually all the enemies)

Scripts/src/Script/Mushdoom.cpp
```diff
@@ -68,6 +68,7 @@ Mushdoom::~Mushdoom()
 
 void Mushdoom::Awake()
 {
+	enemy_type = EnemyType::MUSHDOOM;
```
raulgonzalezupc, line 71: you can put this on the .h since its allways the same and you are on a musdoom class

Scripts/src/Script/EventManager.cpp
```diff
@@ -217,9 +217,13 @@ void EventManager::OnInspector(ImGuiContext* context)
 	ImGui::SetCurrentContext(context);
 	ShowDraggedObjects();
 
-	ImGui::Text("enemies_killed_on_wave: %d", enemies_killed_on_wave);
-	ImGui::Text("Current Battle Event Beated: %d", current_event_beated);
-	ImGui::Checkbox("tutorial_activate", &tutorial_activate);
+	ImGui::DragInt("enemies_killed_on_wave:", &enemies_killed_on_wave);
```
raulgonzalezupc, line 220: Enemies Killed on Wave

PR #89, opened 2020-07-15
Features/vfx

raulgonzalezupc:
VFX implementation on the whole scene:
- VFX particles
- VFX billboards
- VFX spritesheets
- VFX trails
- VFX scripts


Scripts/src/Script/EventManager.cpp
```diff
@@ -85,8 +93,10 @@ void EventManager::Start()
 {
 	player_1 = world->GetPlayer1();
 	player_2 = world->GetPlayer2();
+	dust_roots_VFX->Stop();
+	//audio_source_ambient->PlayEvent("play_forest_ambient");
```
JorxSS, line 97: delete comment

PR #94, opened 2020-07-15
Bugfix/mushdoom spawns

marcorod94:
What's new?
- Fix mushdoom erros that shoots while spwaning


Scripts/src/Script/Mushdoom.cpp
```diff
@@ -502,12 +502,10 @@ void Mushdoom::SpawnJump() const
 	}
 	else
 	{
-		if (!animation->IsOnState("Ground"))
+		/*if (!animation->IsOnState("Ground"))
```
JorxSS, line 505: what is the purpose to have this commented

JorxSS:
There is still fixes to do buuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuut.. it works a "little bit" better
GJ

PR #110, opened 2020-07-27
Feature/attack auto aim

JorxSS:
What's new:

Another iteration in order to improve the combat system.

- Players now can rotate between combo attacks.
- Fixed attacks that had a movement.
- Auto Aim in order to make it easier
- You can disable auto aim from Debug.
- You can spawn a invencible Mushdoom.
- Now you have invincible mode, and god mode to make "ostias como panes", or just don't being hitted.
- Patch Biter crash (Biter team should check that in Biter's next iterations).
- Enemies didn't allow to being hitted while they were playing the hit animation. Now fixed.


Scripts/src/Script/PlayerAttack.cpp
```diff
@@ -573,4 +614,57 @@ void PlayerAttack::DisableTrails()
...
 		}
 	}
-}
\ No newline at end of file
+}
+
+bool PlayerAttack::GetCloserEnemyPosition(float3& closer_position)
+{
+	closer_position = float3::zero;
+	float distance = 6;
```
yalania, line 622: what is this 6?

Scripts/src/Script/DebugModeScript.cpp
```diff
@@ -251,6 +260,12 @@ void DebugModeScript::DrawPlayerOptions()
 	if (ImGui::CollapsingHeader(ICON_FA_USER_INJURED " Player Options"))
 	{
 		ImGui::Checkbox("Invincible mode", &is_player_invincible);
+		ImGui::Separator();		
+		
+		ImGui::Checkbox("God mode", &is_player_god_mode);
+		ImGui::Separator();
```
marcorod94, line 266: This could be quite confusing naming wise because the only difference is a One-hit KO in both modes your life don’t decrease

Scripts/src/Script/EnemyManager.cpp
```diff
@@ -122,6 +122,12 @@ void EnemyManager::SpawnEnemy(EnemyType type, const float3& spawn_position)
 				enemy->collider->UpdateDimensions();
 				enemy->ResetEnemy();
 				++current_number_of_enemies_alive;
+
+				if(invincible)
+				{
+					enemy->invincible = true;
+				}
```
marcorod94, line 129: Wouldn't it be better to assing the value directly to the variable?

Scripts/src/Script/EnemyManager.cpp
```diff
@@ -326,6 +332,7 @@ void EnemyManager::KillAllTheEnemies()
 		if(enemy->is_alive)
 		{
 			enemy->is_alive = false;
+			enemy->invincible = false;
```
marcorod94, line 335: This will be unnecessary

Scripts/src/Script/EnemyManager.cpp
JorxSS: Done

Scripts/src/Script/EnemyManager.cpp
JorxSS: No, because if it's invincible, you can not kill him with the button Kill All the Enemies

marcorod94: Sorry i miss read the line

PR #116, opened 2020-08-03
Feature/light shield

Dartemiss:
What's new?
- Light shield implemented! It stops damage from enemies and makes the lightball rebound. For activate the light shield use "K" on a keyboard or (Triangle/Y) on a controller.
- Knockback is now a parameter by default of the TakeDamage function, and each attack has a different knockback power value.
- Stun has been implemented for the player, when lightshield breaks, the player gets stunned by a certain time.
- Visuals for shield are placeholders (using same as lightorbs) and a stunning effect from the internet.

PD: For trying to balance the lightball rebound I implemented some optional features but test it, please. On the LightBall script, there is a Checkbox called "Use parry for rebound" that will allow rebound if u parry on the correct moment.
Also on the PlayerController there is a checkbox called "Slow with shield" that will slow the player movement speed when using the shield.


Models/Male Character/ShieldEffect.prefab
```diff
@@ -0,0 +1,349 @@
+{
+    "GameObjects": [
```
yalania, line 2: The prefab!

Scripts/src/Script/LightShield.cpp
```diff
@@ -0,0 +1,233 @@
...
+LightShield* LightShieldDLL()
+{
+	LightShield* instance = new LightShield();
+	return instance;
+}
+
+
+LightShield::LightShield()
+{
+
```
yalania, line 28: LightShield() = default on header?

Scripts/src/Script/LightBall.cpp
```diff
@@ -390,6 +404,16 @@ bool LightBall::AlreadyHitted(EnemyController* enemy)
 	return std::find(enemies_hitted.begin(), enemies_hitted.end(), enemy) != enemies_hitted.end();
 }
 
+void LightBall::SwapPlayersTarget()
+{
+	if (world_manager->multiplayer)
+	{
+		std::swap(current_player, next_target);
```
yalania, line 411: std::swap :O niceee

Scripts/src/Script/LightShield.cpp
Dartemiss: yes! done :D

Scripts/src/Script/LightBall.cpp
```diff
@@ -103,7 +104,7 @@ void LightBall::Update()
 
 	if(input_manager->GetGameInput("LightBall", static_cast<PlayerID>(current_player_controller->player - 1)))
 	{
-		if(!moving && CanBeActivated() && current_cooldown <= 0 && current_state == LightBallState::UNNACTIVE)
+		if(!moving && CanBeActivated() && current_cooldown <= 0 && current_state == LightBallState::UNNACTIVE && !current_player_controller->light_shield->active)
```
mariofv, line 107: This condition is starting to be complex and big, could it be possible to encapsulate it in a method?

Scripts/src/Script/LightBall.cpp
```diff
@@ -228,24 +229,36 @@ void LightBall::Update()
...
 				current_cooldown = cooldown;
+				speed = default_speed;
+				maximum_time_to_return = default_maximum_time_to_return;
 
 				//TODO Delete this when OnCollisionEnter implemented
 				enemies_hitted.clear();
 				//Debug
 				transition = false;
 			}
+			else if(((use_parry_for_rebound) ? current_player_controller->light_shield->parry_enable : current_player_controller->light_shield->active)  && ball_distance_magnitude < shield_radius_range)
```
mariofv, line 247: This is quite difficult to read, separate it in different lines. Also having a ternary operator that returns a boolean in the condition is weird, probably the logic could be rewritten in a different way

Scripts/src/Script/PlayerController.cpp
```diff
@@ -126,12 +130,41 @@ void PlayerController::Update()
 		case PlayerState::ALIVE:
 			if(!is_reviving)
 			{
```
mariofv, line 132: This  method is starting to get big, we need to encapsulate behaviours in methods for readability and faster debugging

Scripts/src/Script/LightBall.cpp
Dartemiss: Is was just for debug and letting you test it with parry and without it, once we decide which one we will use ill change the ternary operator by what we choose.

Scripts/src/Script/LightBall.cpp
Dartemiss: Done!

Scripts/src/Script/LightShield.cpp
```diff
@@ -0,0 +1,227 @@
...
+// void LightShield::Load(const Config& config)
+
+// {
+
+// 	exampleUUID = config.GetUInt("ExampleNameforSave", 0);
+
+// 	Script::Load(config);
+
+// }
```
JorxSS, line 227: Delete if not used

Scripts/src/Script/LightShield.h
```diff
@@ -0,0 +1,51 @@
...
+	~LightShield() = default;
+
+	void Awake() override;
+	void Start() override;
+	void Update() override;
+
+	void OnInspector(ImGuiContext*) override;
+	void InitPublicGameObjects();
+	//void Save(Config& config) const override;
+	//void Load(const Config& config) override;
```
JorxSS, line 20: same

Scripts/src/Script/LightShield.cpp
Dartemiss: Done!

PR #118, opened 2020-08-06
Feature/no bullet player movement

JorxSS:
What's new?

* NO MORE BULLET in the movement.
* New STM with the new system implemented by @Dartemiss 

PD: PlayerController.cpp will change after Light Shield merged.
IN ORDER TO TEST THIS:
Have develop up-to-date and delete library.


Scripts/src/Script/PlayerAttack.cpp
```diff
@@ -421,13 +416,13 @@ void PlayerAttack::MakeAttack()
 		if(UpdateAttack(input->button))
 		{
 			CheckDirection();
-
 			animation->ActiveAnimation(current_attack->animation);
+			animation->SetBool("is_attacking", true);
```
Dartemiss, line 420: Why here are you passing true directly instead of changing value of is_attacking to true? Just asking

Scripts/src/Script/PlayerAttack.cpp
```diff
@@ -442,13 +437,14 @@ void PlayerAttack::MakeAttack()
...
 			CheckDirection();
+			
 			if (!debug->disable_autoaim)
 			{
 				AutoAimAttack();
 			}
-			App->engine_log->Log("State changed to: %s", current_attack->animation_state.c_str());
-			App->engine_log->Log("Anim percentatge: %.3f", anim_percentatge);
 			animation->ActiveAnimation(current_attack->animation);
+			animation->SetBool("is_attacking", true);
```
Dartemiss, line 446: Same as before

Scripts/src/Script/PlayerAttack.cpp
```diff
@@ -669,3 +666,22 @@ void PlayerAttack::RotateToAttackDirection()
...
+		break;
+	case Move::HEAVY:
+		animation->SetBool("light_attack", false);
+		animation->SetBool("heavy_attack", true);
+		break;
+	default:
+		animation->SetBool("light_attack", false);
+		animation->SetBool("heavy_attack", false);
+		break;
+	}
```
Dartemiss, line 686: Indent this propertly pls is hurting my heart.

Scripts/src/Script/PlayerMovement.h
```diff
@@ -48,30 +48,31 @@ class PlayerMovement : public Script
...
 
 	bool IsInside(float3& transform) const;
 
 	void InitPublicGameObjects();
 	void Save(Config& config) const override;
 	void Load(const Config& config) override;
 
-	RayHit* GetPlayerIntersection(const LineSegment& ray, bool only_mesh = false);
+	bool IsPlayerColliding(RayHit* collision, const LineSegment& ray, bool only_mesh = false);
+	bool DetectCollisions(const int collision_direction);
```
yalania, line 62: const int doesn't make sense since you are doing a copy here, but you can make do this method and others const

Scripts/src/Script/PlayerMovement.h
```diff
@@ -92,32 +93,49 @@ class PlayerMovement : public Script
...
-	float gravity = 30.f;
-	float vertical_attenuation = 0.5f;
+	//Debug
+	bool is_inside = true;
+	bool visualize_future_aabb = false;
 
-	bool sloppy = false;
-	float angle = 0.f;
-	float threshold = 35;
+	mutable float3 device_coordinates = float3::zero;
```
yalania, line 100: why you do this mutable?

Scripts/src/Script/PlayerMovement.h
JorxSS: I made this in order to be sure to don't modifiy that value, but I cannot make this method const, because I'm modifying some global values in there

Scripts/src/Script/PlayerAttack.cpp
JorxSS: Done

Scripts/src/Script/PlayerMovement.h
JorxSS: because I'm modifying this value in a const method

Scripts/src/Script/PlayerMovement.h
yalania: but if this value is a copy, why are you worry about modifying it?

Scripts/src/Script/PlayerMovement.h
yalania: Yes I mean why are you making it mutable instead of removing the const in the method.

PR #121, opened 2020-08-09
Feature/cinematic camera

Dartemiss:
What's new?
- Cinematic camera has been implemented.
- Different actions for the cinematic camera have been implemented: moveaction, rotateaction, lookataction, movewhilerotateaction, movewhilelookataction, waitaction and setupcameraaction.
- You can define your cinematics on the method CreateCinematics of CinematicCamera and then for playing it call the method PlayCinematic() from CinematicCamera.

As example I created two cinematics, one is played at the begining and the other one when the first battle zone is defeated.


Scripts/src/Script/EventManager.cpp
```diff
@@ -435,11 +436,6 @@ void EventManager::RestartEvents()
 	{
 		battle->RestartEvent();
 	}
-	for (auto & root : root_obstacles)
-	{
-		root->SetEnabled(true);
-		root->children[0]->GetComponent(Component::ComponentType::MESH_RENDERER)->active = false;
-	}
```
yalania, line 442: this was unused?

Scripts/src/Script/CinematicCamera.h
```diff
@@ -0,0 +1,46 @@
...
+	void Start() override;
+	void Update() override;
+
+	void OnInspector(ImGuiContext*) override;
+	void InitPublicGameObjects();
+	void SetMainCamera(bool cinematic) const;
+	void PlayCinematic(const std::vector<CinematicAction*>& cinematic);
+	void PlayCinematic(size_t index);
+
+	void CreateCinematics();
```
yalania, line 28: Can you mark as private the methods that are not from the API? so it is easier to use?

Scripts/src/Script/CinematicCamera.cpp
```diff
@@ -0,0 +1,427 @@
...
+			finished = true;
+			return;
+		}
+
+		Quat rot = Quat::Slerp(camera->transform.GetRotation(), look_quat, delta_time * 0.004f);
+		camera->transform.SetRotation(rot);
+	}
+};
+
+class MoveWhileRotateAction : public CinematicAction
```
yalania, line 129: Instead of creating a MoveWhileRotateAction  can it be played to actions one RotateAction  and one MoveAction at the same time?

Scripts/src/Script/CinematicCamera.cpp
```diff
@@ -0,0 +1,427 @@
...
+class CinematicAction
+{
+public:
+	bool finished = false;
+	float duration = 0.f; //ms
+	float current_time = 0.f; //ms
+	GameObject* camera = nullptr;
+
+	virtual void Update(float delta_time) = 0;
+};
```
yalania, line 31: if this class is define in the cpp you can't access them in others cpp files as you can't include it.

Scripts/src/Script/CinematicCamera.cpp
Dartemiss: No, if you want that moves and rotates at the same time you have to use MoveWhileRotate. I recomend using MoveWhileLookAt it works more smooth

Scripts/src/Script/CinematicCamera.cpp
Dartemiss: Yes i know at the beginning was everything on Cinematics.h but i had tons of errors and i moved the code here. I could move the code to the .h but not other class should use this classes

Scripts/src/Script/EventManager.cpp
Dartemiss: I haven't touched that code wtf I am going to check how is on develop.

Scripts/src/Script/CinematicCamera.h
Dartemiss: Yes! Done!

Scripts/src/Script/CinematicCamera.cpp
yalania: then void PlayCinematic(const std::vector<CinematicAction*>& cinematic); can be properly use, as you can't create any CinematicAction from outside ?

Dartemiss: True, that was my first aproach but I think ill delete that option

Dartemiss: Delete it!

Scripts/src/Script/EventManager.cpp
```diff
@@ -483,6 +479,8 @@ void EventManager::EndEvent()
 			ambient_audio->PlayEvent("Stop_Corrupted_Ambience");
 			ambient_audio->PlayEvent("Play_Forest_Ambience");
 
+			//Trigger cinematic for testing
+			cinematic_camera->PlayCinematic(1);
```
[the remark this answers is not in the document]
Dartemiss: The cinematic of index 1. They are created on the CinematicCamera CreateCinematics()

Scripts/src/Script/TerrainDamage.h
```diff
@@ -2,12 +2,11 @@
 #define  __TERRAINDAMAGE_H__
 
 #include "Script.h"
+#include "PlayerController.h"
+#include "WorldManager.h"
```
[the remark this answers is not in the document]
Dartemiss: Done!

Scripts/src/Script/CinematicCamera.cpp
```diff
@@ -0,0 +1,421 @@
...
+public:
+
+	RotateAction(const float3& rotation_degrees, const Quat& current_rotation, float duration, GameObject* camera)
+	{
+		this->rotation_degrees = rotation_degrees;
+		this->duration = duration;
+		this->current_rotation = current_rotation;
+		this->camera = camera;
+
+		float3 current_direction = camera->transform.GetFrontVector() + camera->transform.GetGlobalTranslation();
```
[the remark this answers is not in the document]
Dartemiss: It is an indirect for implement it? XDDD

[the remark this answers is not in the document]
mariofv: image

mariofv (approved):

PR #122, opened 2020-08-13
Feature/ally

yalania:
- Dialogues for ally
- Lightball recieve from ally
- Move around the scene

Note: Right now you can skip recieving the lightball, I will continue developing it once we have this merge.


Scripts/src/Script/AllyInteractionsScript.cpp
```diff
@@ -0,0 +1,170 @@
...
+	{
+		can_start_next_dialogue = false;
+		camera->active = false;
+	}
+}
+
+//Copied from mushdoom
+void AllyInteractionsScript::RotateTowardsPlayer() const
+{
+	float3 desired_direction = player_1->player_go->transform.GetGlobalTranslation() - owner->parent->transform.GetGlobalTranslation();
```
Dartemiss, line 165: It only looks to the player1? Maybe could Look into the closer player instead

yalania: in most multi games only the player one can trigger conversations etc, so I just used player1

Scripts/src/Script/LightBall.cpp
```diff
@@ -102,7 +93,7 @@ void LightBall::Update()
 		return;
 	}
 
-	if(input_manager->GetGameInput("LightBall", static_cast<PlayerID>(current_player_controller->player - 1)))
+	if(!mAllyHasLightBall && input_manager->GetGameInput("LightBall", static_cast<PlayerID>(current_player_controller->player - 1)))
```
JorxSS, line 96: cs here

Scripts/src/Script/LightBall.h
```diff
@@ -158,6 +159,9 @@ class LightBall : public Script
 
 	//Delete when the animations and STM work better
 	float frame_counter = 0.f;
+
+	//Ally LightBall Tutorial
+	bool mAllyHasLightBall = true;
```
JorxSS, line 164: cs here with this variable

Scripts/src/Script/Script.cpp
```diff
@@ -91,5 +91,6 @@ void Script::ShowDraggedObjects()
 
 Script::~Script()
 {
+	//delete panel;
```
JorxSS, line 94: ???

yalania: I have seen there are a lot of memory leaks in the gameplay project, this an example but need to add ENGINE_API to the destructor, as you are going to test it with master branch I left it commented so you don't have linking issues

Dartemiss: Which memory leaks did you find?

yalania: just do a search for = new and you will se most of the allocated stuff is never deleted, right now I don't remember all of them

Dartemiss: wow u are right, all the instances and the states of the STM of enemies.

JorxSS (approved):
Sorry for the delay
image

Dartemiss (approved):
GJ Anabel! :D

PR #123, opened 2020-08-28
Feature/juicy damage indicators

mariofv:
What's new:
- `DamageIndicator` trajectory is parabloic.
- `DamageIndicator` trajectory is determined by hit direction.
- `DamageIndicator` font is bolder.
- `DamageIndicator` color changes with power of the hit.
- `DamageIndicator` size changes with power of the hit.


JorxSS (approved):
This is pretty cool! GJ!!

image

Scripts/src/Script/DamageIndicator.cpp
```diff
@@ -41,36 +43,82 @@ void DamageIndicator::Start()
...
 }
 
 bool DamageIndicator::IsAlive() const
 {
-	return alive;
+	return state != DamageIndicatorState::DEAD;
 }
 
-void DamageIndicator::Spawn(int number, float2 position)
+void DamageIndicator::Spawn(int number, float power, float2 position, const float2& direction)
```
Dartemiss, line 102: position could be const ref?

Scripts/src/Script/DamageIndicatorSpawner.cpp
```diff
@@ -14,9 +16,12 @@
 
 
 #include "DamageIndicator.h"
+#include "PlayerAttack.h"
 
 #include "imgui.h"
 
+const float MAX_DAMAGE = 55.f;
```
Dartemiss, line 23: Max damage isn't 55

JorxSS: It is, and I told him that value, but now that I'm thinking, if we are going to do power ups, that maximum damage will increase

Dartemiss (approved):
Very good job man! I love the result. With a good sound effect depending on attack/damage too will be amazing.

Scripts/src/Script/DamageIndicatorSpawner.cpp
Dartemiss: Lightball can deal until 71 damage, so it isn't

Scripts/src/Script/DamageIndicator.cpp
mariofv: No, because I'm modifying position later in the method

PR #124, opened 2020-08-30
Feature/ally2

yalania:
- Ally has now two forms, humanoid (only for intro) and lightball
- Added some triggers for ally behaviour, to force the player to tak with the ally and to force conversations on lightball mode

Need to add animations, visual effects and better texts. There is a bug if you go into the trigger starting the jump (animation get stuck) any suggestion?


Dartemiss (commented):
GJ, works nice! I would put a cinematic at the beginning pointing to the ally, emphasising that you have to talk to him. Also, maybe on the dialogues or somehow it can tell you something about the mushdooms .
About when it becomes the lightball, it should explain that he will help you with some cool powers, and explains you how to use them (lightball and shield). Also, it should be implemented that the ally talks to you from your insides (just a dialog on ui, with his face).

JorxSS (changes requested):
I found some problems with the Ally.
- Ally needs to have the GO Ally_positions in the scene and if not, crash. (And suggestion, change the GO name to Ally Positions without underscore)
- As we want to have level selection, we cannot use the lightball if we are in another level that isn't the first one.

Dartemiss:
Second point that jordi commented is fixed on our branch of boss for being able of using it.
- "As we want to have level selection, we cannot use the lightball if we are in another level that isn't the first one."

yalania:
- On Ally controller if you chose Ally_is_lightball (this setting is saved with the scene) you can use the lightball

JorxSS (approved):
LGTM

Dartemiss (approved):
GJ Anabel! Just what we have commented making play of the vfx!

PR #125, opened 2020-09-02
fix/movement

JorxSS:
What's new:

Some fixes in the movement.


Scripts/src/Script/AllyInteractionsScript.cpp
```diff
@@ -82,7 +82,7 @@ void AllyInteractionsScript::Update()
 		{
 			camera->active = true;
 			ActivateDialogue();
-			App->engine_log->Log("Talking with Ally");
+			//App->engine_log->Log("Talking with Ally");
```
Dartemiss, line 85: Why is this comented?

Dartemiss (approved):
GJ my pana, works amazing! Just the debug comments i don't understand why are they there.

mariofv (approved):
Some bugs are fixed, but this happened in the rocks 
https://i.gyazo.com/d703918327f4c088471e914630e0a2e7.gif

PR #127, opened 2020-09-07
Feature/clean dependencies

JorxSS:
What's new:

* Fixed again Playground
* Cleaning Environment VFX dependencies.
* Fixed water zone damage again, thanks @Dartemiss.


Scripts/src/Script/EventManager.cpp
```diff
@@ -447,7 +405,11 @@ void EventManager::RestartEvents()
 	{
 		battle->RestartEvent();
 	}
-
+	for (auto & root : root_obstacles)
```
Dartemiss, line 408: CS

Dartemiss (changes requested):
GJ! Works more or less everything but! It crashes when finishing the game on level1 entering the cave (see image below)

EventManager
imagen

Also you already know but particles takes too much time to spawn.

Scripts/src/Script/EventManager.cpp
JorxSS: Done

JorxSS:
@Dartemiss already fixed the feature/bug now cannot happen sorry mi pana

Dartemiss (approved):

Pacasasgar:
I found this bug while testing this version:
water_death_bug

As you mention in the description, water zone has been fixed. Is this another unkown bug? Player do not die but can not move either in the gif's state.

JorxSS:
> I found this bug while testing this version:
> water_death_bug
> 
> As you mention in the description, water zone has been fixed. Is this another unkown bug? Player do not die but can not move either in the gif's state.

known, will be fixed with the new animations

Pacasasgar:
> > I found this bug while testing this version:
> > water_death_bug
> > As you mention in the description, water zone has been fixed. Is this another unkown bug? Player do not die but can not move either in the gif's state.
> 
> known, will be fixed with the new animations

Great then!

Pacasasgar (approved):
Nice job Jordi!

PR #135, opened 2020-09-11
Audio Manager

Pacasasgar:
This PR contains the following changes:
-A new important script has been created (Audio Manager). It manages all the audio events and simplifies all audio_sources in one place. From now on, to play audio events we should use the audio sources from Audio Manager.
-EventManager now uses audio sources from AudioManager for 2D scene global audios (only using one object, yay!).
-Removed a lot of Scene-dependent code from EventManager. Now it works following the logic from Jordi's EnvironmentVFX - zones.
-Environment 3D sounds are now fixed. They also change from corrupted/uncorrupted if player beats every zone.

It also needs this PR to be merged into LittleOrionEngine: 
https://github.com/OnionGalaxy/LittleOrionEngine/pull/275


Scripts/src/Script/AudioManager.cpp
```diff
@@ -0,0 +1,99 @@
...
+	//Necessary to be able to write with imgui
+	ImGui::SetCurrentContext(context);
+	ShowDraggedObjects();
+}
+
+//Use this for linking JUST GO automatically 
+void AudioManager::InitPublicGameObjects()
+{
+	//IMPORTANT, public gameobjects, name_gameobjects and go_uuids MUST have same size
+	for (int i = 0; i < public_gameobjects.size(); ++i)
```
RickyVimon, line 60: better use unsignet int

RickyVimon (approved):
Make the changes mentioned above and should be fine. Next thing to do is to reset the audiosources when you die and respawn.

Dartemiss (approved):

PR #136, opened 2020-09-11
Fix credits menu

luismoyano:
credits

Credits menu improved


luismoyano:
Even thoug under Lead designer it says Mario's name, it's correct on the commit, I noticed the error while recording the gif

imarcuello (commented):
On a second look I also saw that the title of credits it's a little bit deformed in the vertical axis. 
The rest with the gif modification it's ok.

JorxSS:
@luismoyano I don't know if you currently fixed but, why "Producer" is not in caps? and also, why is not there our Lead Artist?

luismoyano:
@JorxSS True, my mistake, its fixed now

vidalmelero (approved):

JorxSS (changes requested):
image

imarcuello (approved):
I've seen the final outcome, with everything correcter it looks good!

JorxSS (approved):

PR #140, opened 2020-09-13
Better scaling for character selection ui

luismoyano:
I reorganized things more evenly, modified the background so it doesnt feel out of place when unscaled and also fixed the deformations on the icons' texts

image


Dartemiss (approved):
GJ! Works fine!

mariofv (approved):
Good job!

PR #148, opened 2020-09-21
Change some materials to implement bloom

mariofv:


raulgonzalezupc (approved):
Keep the scene of the alpha 2 and we have it!

PR #149, opened 2020-09-21
Feature/player effects

raulgonzalezupc:
Prefabs of :
 - Player
 - Mushdoom
 - Biter
 - Necroplant

VFX of hit and attacks


Dartemiss (changes requested):
GJ! There are some bugs on the effects, also you modified the level1 scene and now has a necroplant at the beginning, also the light attacks vfx are on the oposite (the first ones are big and the last hit is little). And I assume that the shield is placeholder so it's okay.

Dartemiss (approved):
LGTM

mariofv (approved):
They look great, GJ

PR #156, opened 2020-09-25
Feature/improve water material

yalania:
- Water material + texture update
- added vfx to watefall to be reviewed (maybe new spritesheet?)
- added a spotlight focusing the water.


raulgonzalezupc:
movement too linear, we need some more flow
particles on the hit at the waterfall can't be grey, the water is not transparent as usuall, also we need more particles on the end I think. 
Tiling on the water is too much, I think we need to reduce it a little bit

Dartemiss (changes requested):
I would try to make the water final instead of improving little by little as this is the last big delivery. It looks better but the movement is too lineal.

yalania:
mmm I have tried deforming he mesh, but there are threeproblems first is that there are some lotus that have not movement, so adding more movement at the water is weird, second is that the waterfall and the river are the same mesh so the same shader and movement is applied to them, third this mesh doesn't have many triangles so we would need to apply some extra shading to do a good movement.

I can may add an extra normal maps flow on top of it so there are like two different flows or try with a different normal map.

PR #157, opened 2020-09-26
fix vfx stuck after talking with ally

yalania:
Issue was because I was doing a disable of all the player GO, so when enabling ti again all the billboards were enable by default. Now I'm only disable the component mesh renderer


Scripts/src/Script/AllyInteractionsScript.h
```diff
@@ -95,8 +96,8 @@ class AllyInteractionsScript : public Script
 	bool current_lightball_speed = 0.005f;
 	float time_passed = 0.0f;
 	//Camera view
-	GameObject* player_1_mesh = nullptr;
-	GameObject* player_2_mesh = nullptr;
+	Component* player_1_mesh = nullptr;
```
raulgonzalezupc, line 99: why not ComponentMeshRenderer?

raulgonzalezupc (approved):
Tested and seems that works

Scripts/src/Script/AllyInteractionsScript.h
yalania: because I don't need to know specific things of componentMeshRenderer and I can avoid a casting.

marcorod94 (commented):
Looks nice but @yalania I don't if you noticed the same thing happens when you revive a player could you look into it?

yalania:
oki

marcorod94 (approved):
Nice Job

PR #158, opened 2020-09-26
Bug 42 fix + not allow shielding while dead

Dartemiss:
Bug 42 fix: Shield gets automatically activated while reviving a dead player
Now the player cannot shield while he is dead.


Scripts/src/Script/PlayerController.cpp
```diff
@@ -199,6 +199,8 @@ void PlayerController::OnInspector(ImGuiContext* context)
 	ImGui::DragFloat("Total Special Points", &total_special_points);
 
 	ImGui::Checkbox("Is Alive", &is_alive);
+	ImGui::Checkbox("Is Reviving", &is_reviving);
+	ImGui::Checkbox("Is Dashing", &is_dashing);
```
raulgonzalezupc, line 203: I assume you putted this here because it wasn't before right? you don't use this variable on the new code nowhere

raulgonzalezupc (approved):
Tested with both players and works perfect!

Scripts/src/Script/PlayerController.cpp
Dartemiss: Was for testing yeah! And more info on inspector!

PR #162, opened 2020-09-27
Fix/ally

yalania:
- Fix dialogues to avoid cuts
- When teleport if player doesn't have the lightball it will be throw from the ally
- Require 3 consecutive lightball bounces to pass the tutorial (maybe to change 3 in total? I'm not sure what is best)
- Don't stun player during tutorial to avoid bugs


Dartemiss (changes requested):
The text don't cut anymore gj! And 3 rebounds work too!
But I found several bugs:
- When teleporting to zone two if u go back a bit u teleport to a bugged tutorial when tutorial should be skipped. https://gyazo.com/e1f3802383244563253122c20d2887d8
- Shield tutorial on multiplayer is broken, Player2 is not able to use shield to rebound the ball and the UI don't match with which players has the de ball so is very confusing.
- If you broke the shield on battle zone 1 and you go to the shield tutorial you can't use shield until is recovered.

yalania:
for the teleport you have to wait a little bit because the litghball takes some time to reach the player.
For the shield and durability on tutorial I'm not sure what should I be doing, do I just ignore durability on the tutorial? maybe in that case the player would be confuse if after the tutorial it can be break?

Dartemiss:
It happens even with you wait for the lightball on the teleport, I would reestart the shield on tutorial anyways I think they want to change that the shield can be broke

Dartemiss (approved):
GJ Works fine now! :D

mariofv (approved):
LGTM, text looks good!

PR #163, opened 2020-09-28
Fix/bug 43 restart boss fight

Dartemiss:
Fix bug 43.
Now when players respawns is invulnerable for 3 seconds and boss remains in idle for 4 seconds.

Also fixes bug 8: Level 1: When you die in the river, you will lose 2 lifes 
instead of one

PD: Boss scene is changed on this branch


raulgonzalezupc:
When you die on the boss part, it does not go to iddle for 4 seconds. River part works fine!

Dartemiss:
Did the boss damaged you while respawning? You are invulnerable on that time.

Scripts/src/Script/BossController.cpp
```diff
@@ -33,6 +33,17 @@
...
+{
+	const float min_shadow_scale = 0.050f;
+	const float max_shadow_scale = 0.150f;
+	const float remaining_crystal_time_alive = 1000.f;
+	const float max_remaining_rock_time_alive = 1000.f;
+	const float min_remaining_rock_time_alive = 200.f;
+	const float time_stunned = 2000.f;
+	const float time_waiting_for_spawning = 4000.f;
+}
+
```
raulgonzalezupc, line 46: there are variables here that you don't use in all the code

Dartemiss: which ones?

mariofv (approved):
LGTM

raulgonzalezupc:
> Did the boss damaged you while respawning? You are invulnerable on that time.

It did not, but it didn't did the iddle animation for 4 sec as you said

Misarte (approved):
I think for me it works, just the idling happens bit later than I expect it.

PR #164, opened 2020-09-29
Fix bug 37

Dartemiss:
Fix bug where the player could use dash multiple times on the air, flying


yalania (approved):

Pacasasgar (approved):
Bugfix tested and correctly solved. Good Job Enrique!

PR #165, opened 2020-09-29
Break line for ComponentText

marcorod94:
Now we could use \\n to break lines on dialogue

Use it with the same branch on the LittleOrion Engine


yalania:
There is a weird animation when there is a jump, I don't know if it is because of the dialogue manager o the component implementation, could you check it?

Dartemiss (commented):
imagen

Dartemiss:
imagen

Also here is cut!

marcorod94:
The last image isnt loading for future texts we should add \n to break the line

Dartemiss:
Talk with @imarcuello for checking how the UI text should be please :D

Dartemiss:
imagen

Now this happens for some reason

Dartemiss:
imagen

Here too, there is like a weird tab

yalania (approved):

Dartemiss:
This is not used as the \n doesnt work propertly and UI team is working on that. I close it

PR #166, opened 2020-09-29
Feature/lightmaps

mariofv:
What's new:
- Apply Lightmaps to `Level1` materials.
- Corrected far distance to `Level1` `MainCamera`.
- Set shadow parameters in different `Level1` entities.

What has to be done in each Level:
- Reduce `MainCamera` far plane distance to 100 or so.
- Enable Shadows (`Config`->`Render`->`Shadows`->`Enable Shadows` and `Save Scene`)
- Set `ShadowCaster` for player mesh and biters (or anything that we want casting shadows).
- Set `ShadowReceiver` for ground and walls of the level.
- Set Lightmap texture when neccessary.
- Put an angle to directional light that makes shadows look good.


Dartemiss (approved):
LGTM

yalania (commented):
shadows are a little bit pixelated
image

A part from that as I said in the engine branch, is there anyway to make the shadow softer?
(I enabled shadow on trees to see the effect don't worry)

mariofv:
> shadows are a little bit pixelated
> image
> 
> A part from that as I said in the engine branch, is there anyway to make the shadow softer?
> (I enabled shadow on trees to see the effect don't worry)

Yes, this is expected. I already pushed in engine to render ambient color in shadowed areas to make them softer. It's pixelated due to the resoultion of the light frustum. 

Few ways to fix pixelation are changing the shadows to cascade mapping and modifying far plane from main camera, or putting closer objects are shadow caster

PR #167, opened 2020-09-30
Fix/lightball improvements

JorxSS:
What's new:
- Fixed Bug 18: Lightball when cast appear in the other player.
- Fixed Bug 34: Lightball always going on the plane.
- Fixed Bug 47: If player that doesn't have the ball dies, when revives
cannot receive the ball anymore.
- Fixed Bug 48: When the players that have it dies, goes to the other one
but so small that the user can't barely see it.

Lightball Auto Aim will come soon...


yalania (approved):

Dartemiss (approved):
LGTM

PR #168, opened 2020-10-01
Feature/hit sfx per enemy and attack

Dartemiss:
What's new?
- Implemented HitSFX on enemy controller and its derived classes
- Now you can choose a different hit sfx for each attack of the player and each enemy.

PD: Biter code is commented as it doesn't have audio source yet because there is no biter soundbank


Scripts/src/Script/MushdoomStateDie.cpp
```diff
@@ -26,7 +26,7 @@ void MushdoomStateDie::OnStateEnter()
 		enemy->animation->ActiveAnimation("die");
 	}
 
-	enemy->audio_source->PlayEvent("play_receive_hit");
+	//enemy->audio_source->PlayEvent("play_receive_hit");
```
Pacasasgar, line 29: Remove this if not useful anymore

Pacasasgar (approved):
Tested and looking (hearing) good! With this implementation it will be much easier to try different hit sounds depending on the enemy / attack. Left a simple comment to wath it out but overall gj!

RickyVimon (approved):
Gj man! LGTM

PR #169, opened 2020-10-01
Feature/footsteps dust new aproach

raulgonzalezupc:
New footsteps logic:
    - Now the footsteps starts to play when the player starts walking instead to emit each time that the player walks
    - Texture changed
    - Opacity over time
    - Color changed (visuals of the footsteps in general changed)


Scripts/src/Script/PlayerMovement.cpp
```diff
@@ -187,7 +187,18 @@ void PlayerMovement::Move(int player, bool is_stunned)
 		is_jumping = false;
 		animation->SetBool("is_jumping", is_jumping);
 	}
-
+	if (!animation->IsOnState("Idle") && !animation->IsOnState("Dead") && !animation->IsOnState("Stunned") && !animation->IsOnState("Falling") && !animation->IsOnState("Jump"))
```
Dartemiss, line 190: So if player is attacking, or reviving particles will appear? Why dont put IsOnState("Run")?

Dartemiss (approved):

yalania:
when you stop walking the footsteps dissapear, it feels a little bit strange, is there some way we can fix it?

Scripts/src/Script/PlayerMovement.cpp
raulgonzalezupc: I thought that we also wanted it when attacking and dashing

raulgonzalezupc:
> when you stop walking the footsteps dissapear, it feels a little bit strange, is there some way we can fix it?

I tried several times to fix it, but is super weird because I do the stop of the PS, and this should work, but it does not some times :/

yalania (approved):

PR #170, opened 2020-10-01
Biter improvements and bugfixing

Pacasasgar:
This PR solves bug 32 and bug 33.
-Added a new cooldown between special attacks. Now after charging or exploding, biter will not explode or charge (respectively) immediately and instead will keep chasing or basic attacking you. This avoids an explosion attack immediately after a charge attack or the reverse of that. This way, player will have a better feeling when the biter is just chasing him/her or doing an special attack.
-When a biter spawns, it will need to wait its special attack cooldowns in order to do them. This will avoid explosions and charges when Biter is still too high and unreachable from the players.
-Biter height offset modified a bit.


Dartemiss (approved):
LGTM

PR #171, opened 2020-10-01
Feature/skip cinematics

marcorod94:
Skip cinematics using punch button J\B

Known hiccup the skip UI may not appear on the first run sometimes


Scripts/src/Script/CinematicCamera.cpp
```diff
@@ -303,11 +310,23 @@ void CinematicCamera::Update()
...
 		//Retrieve next element
 		current_action = cinematic_queue.front();
+		ui_manager->ShowSkipButton();
+		player_1->player_controller->player_attack->is_attack_blocked = true;
 	}
 
 	//If action we call update
 	if(current_action)
 	{
+		if (input_manager->GetGameInputDown("Punch", static_cast<PlayerID>(player_1->player_controller->player - 1)))
```
Dartemiss, line 320: Only player1 can skil cinematics?

Scripts/src/Script/UIManager.cpp
```diff
@@ -399,9 +399,41 @@ void UIManager::InitPublicGameObjects()
...
+	selected_skip_button->SetEnabled(true);
+}
+
+void UIManager::HideSkipButton()
+{
+	if (selected_skip_button != nullptr)
+	{
+		selected_skip_button->SetEnabled(false);
+	}
+}
```
Dartemiss, line 427: const

Scripts/src/Script/UIManager.cpp
```diff
@@ -399,9 +399,41 @@ void UIManager::InitPublicGameObjects()
...
+{
+	if (input_manager->total_game_controllers == 0)
+	{
+		return skip_button_keyboard;
+	}
+	else
+	{
+		return skip_button_controller;
+	}
+}
```
Dartemiss, line 439: const

Dartemiss (changes requested):
Works fine but I haven't seen any UI at all, even restarting. Also some const correctness

Scripts/src/Script/CinematicCamera.cpp
marcorod94: Yes only the player1 could do that

yalania (commented):
I tried it but it doesn't work for me, the skip ui stay on the middle of the screen (I tried J in keyboard and B in controller)

yalania (approved):
now it works fine

yalania (changes requested):
I tried second cinematic, and when you skip it continues playing at the moment it was skip.

marcorod94:
Everything done please check it

Dartemiss:
This uses old cinematic system so I close this.

PR #173, opened 2020-10-03
Bug 41: Biter did a bite attack while doing the dying explosion

Pacasasgar:
This PR solves bug 41: Biter did a bite attack while doing the dying explosion
-Refactored 'TakeDamage' function from PlayerController. Now it takes a Billboard instead of an EnemyType in order to avoid using the same VFX for every attack of the same enemy (this was causing bug 41).


Scripts/src/Script/PlayerController.cpp
```diff
@@ -231,34 +231,20 @@ void PlayerController::Load(const Config& config)
...
 { 
 	TakeDamage(damage);
 
 	if (light_shield->IsActive())
 	{
 		return;
 	}
-	switch (enemy_type)
+
+	if (vfx)
```
mariofv, line 243: For better reading is better to write `if (vfx != nullptr)`

Scripts/src/Script/EnemyController.h
```diff
@@ -48,10 +48,11 @@ class EnemyController : public Script
 	bool PlayerInAttackRange() const;
 	bool PlayerInAttackArea(const float3& area, const float& damage_radius) const;
 	bool PlayerInRange(float range) const;
-	void DealDamageToPlayersInArea(const float3& area, const float& damage_radius, float damage, bool& player_1_hit, bool& player_2_hit) const;
+	void DealDamageToPlayersInArea(const float3& area, const float& damage_radius, float damage, bool& player_1_hit, bool& player_2_hit, const bool& is_explosion) const;
```
mariofv, line 51: It feels strange to have the parameter `is_explosion` in the class `EnemyController`

Scripts/src/Script/EnemyController.cpp
```diff
@@ -170,6 +170,27 @@ float3 EnemyController::CollisionAvoidance(const float3& velocity) const
 	return avoidance_force;
 }
 
+ComponentBillboard* EnemyController::GetAttackVFX(const Player* player, const bool& is_explosion) const
```
mariofv, line 173: This definition of the method could be done in each enemy class as an `overwrite`, instead of seperating here the cases with `enemy_type`

mariofv (approved):
Effectively this solves the bug, but the whole code regarding hit vfx is a bit messy with mixed logic. (I know that this was not done by you). Because of that, adding new effects is a bit tedious and is really prone to bugs as the one that you solved.

I suggest to rethink how we spawn VFX, and to implement a new way for doing this. I have more or less an idea in mind and it's too long to be written. Let me know if you want to do it!

Pacasasgar:
> Effectively this solves the bug, but the whole code regarding hit vfx is a bit messy with mixed logic. (I know that this was not done by you). Because of that, adding new effects is a bit tedious and is really prone to bugs as the one that you solved.
> 
> I suggest to rethink how we spawn VFX, and to implement a new way for doing this. I have more or less an idea in mind and it's too long to be written. Let me know if you want to do it!

I totally agree with you. I tried a quick fix but had to 'hardcode' a lot and also had the feeling that we need a deeper look into this and re-do a lot. I put Jordi as reviewer and explained him in private more or less the same you told me. I was asking him for suggestions and I will be glad to hear yours. I want for sure do this, so let's meet so you could explain me the solution/refactor.

Scripts/src/Script/EnemyController.cpp
Pacasasgar: You are right. However, right now we only have two different VFX hit effects for our enemies. In fact, I only check in that function that the enemy is the biter and the attack is the explosion, for all the rest cases use the other (default) Hit effect. This should be modified anyway in the VFX refactor we mentioned.

Scripts/src/Script/EnemyController.h
Pacasasgar: Right. This should be our main issue of concern for the VFX refactor.

Scripts/src/Script/PlayerController.cpp
Pacasasgar: Done!

Dartemiss (approved):
LGTM

raulgonzalezupc (approved):
nice work!

PR #175, opened 2020-10-08
Fix/bug multiplayer movements stuck

Misarte:
This bug is bug No 11, I apologize for forgetting the number in the name.

I noticed that the problem was not while pausing the menu but with the IsInside function in PlayerMovement. Sometimes one of the 2 players could get out of the frustum and due to the minimum distance between 2 players noone of the 2 players could move. 

I added a float4 enlarging the area where IsInside is true and I could not reproduce the bug anymore, I tested around the levels and nothing new appeared, so I guess it's fixed.


Scripts/src/Script/PlayerMovement.cpp
```diff
@@ -1,653 +1,653 @@
...
+	device_coordinates = clip_coordinates.xyz() / clip_coordinates.w;
+
+	if (device_coordinates.x < 0.95 && device_coordinates.x > -0.95 && device_coordinates.y < 0.95)
+	{
+		inside = true;
+	}
+	if(other_player != nullptr && world->multiplayer)
+	{
+		float4 position_float4_other_player = float4(other_player->transform.GetGlobalTranslation() - distance, 1.f);
+		float4 clip_coordinates_other_player = game_camera->GetClipMatrix() * position_float4_other_player + float4{0.1f, 0.1f, 0.2f, 0.f};
```
Pacasasgar, line 381: I have tested it and I couldn't reproduce the bug either. This float4 you added seems a bit small but in fact it works well. Good job! We should keep an eye on this just in case anyone find the bug again and thinks it is a new one.

Pacasasgar (approved):

Dartemiss (changes requested):
I could get stucked easily on the first battle zone of leve2. Also I merged develop on your branch because otherwise level2 crashed.

https://gyazo.com/6ab520514f41120ef08b09f8c62f4372

Pacasasgar:
> I could get stucked easily on the first battle zone of leve2. Also I merged develop on your branch because otherwise level2 crashed.
> 
> https://gyazo.com/6ab520514f41120ef08b09f8c62f4372

I did not even try level 2 because I know level 2 is being changed and work in progress. But you are right, it should be working at least in the actual level 2, my bad not testing it.

Misarte:
I made changes and tried again Level2. It seems like its working now. Could you guys throw another look?

Dartemiss (approved):
Now works cool

PR #176, opened 2020-10-08
Feature/powerups

JordiRomagosa:
Changes made:

- Unified the enemy drops into a single class for light orbs (healing) and powerups.
- Implemented two powerups: increased attack and invulnerability.
- Powerups have a greater chance to drop as more enemies are killed, but the chances are never 0% nor 100%.
- Added little feedback (vfx) for the players to know they have the powerup active.

Still waiting for the UI feedback that is not implemented yet.


Scripts/src/Script/LightOrb.cpp
```diff
@@ -62,180 +63,64 @@ void LightOrb::Awake()
...
 
-// Use this for initialization
-void LightOrb::Start()
-{
-	player_1 = world->GetPlayer1();
-	player_2 = world->GetPlayer2();
-}
-
 // Update is called once per frame
 void LightOrb::Update()
```
Dartemiss, line 67: Why is LightOrb class needed now?

Scripts/src/Script/LightOrb.h
```diff
@@ -12,22 +12,14 @@ class WorldManager;
...
-		TRAVELLING,
-		DEAD
-	};
-
 	LightOrb();
 	~LightOrb() = default;
 
 	void Awake() override;
-	void Start() override;
+	//void Start() override;
```
Dartemiss, line 22: If not used delete it

Scripts/src/Script/PlayerController.cpp
```diff
@@ -633,13 +636,57 @@ bool PlayerController::CheckDashBehaviour()
...
+	time_remaining_attack_powerup = 0;
+	time_remaining_invulnerable_powerup = 0;
+
+	powerup_attack_VFX->Stop();
+	powerup_invul_VFX->Stop();
+}
+
+void PlayerController::UpdatePowerups()
+{
+	float delta_time = App->time->delta_time / 1000;
```
Dartemiss, line 651: 1000.f

Scripts/src/Script/PlayerController.cpp
```diff
@@ -633,13 +636,57 @@ bool PlayerController::CheckDashBehaviour()
...
+
+	if (time_remaining_attack_powerup != 0)
```
Dartemiss, line 653: .f

Scripts/src/Script/PowerUp.cpp
```diff
@@ -0,0 +1,214 @@
...
+				{
+					shine_texture_invul = static_cast<ComponentBillboard*>(ch->GetComponent(Component::ComponentType::BILLBOARD));
+				}
+				else if (ch->tag == "exterior_texture_light_orb_VFX")
+				{
+					exterior_texture_invul = static_cast<ComponentBillboard*>(ch->GetComponent(Component::ComponentType::BILLBOARD));
+				}
+			}
+		}
+	}
```
Dartemiss, line 76: Probably is more readable to have all of this on a separate method

Scripts/src/Script/PowerUp.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+	switch (powerup_type)
+	{
+	case PowerUpType::ATTACK:
+		player_to_deliver_controller->time_remaining_attack_powerup += attack_powerup_duration;
+		player_to_deliver_controller->attack_powerup_bonus = 1 + attack_powerup_bonus;
+		break;
+
+	case PowerUpType::INVULNERABLE:
+		player_to_deliver_controller->time_remaining_invulnerable_powerup += invulnerable_powerup_duration;
```
Dartemiss, line 130: This means that if a player gets 3 powerups at the same time he will have the bonus the triple of the time right?

Scripts/src/Script/PowerUp.cpp
```diff
@@ -0,0 +1,214 @@
...
+	case EnemyDropLifeState::WAITING:
+		audio_source->PlayEvent("play_orbs_dropping");
+		break;
+
+	case EnemyDropLifeState::TRAVELLING:
+		break;
+	}
+}
+
+void PowerUp::PickupActivatePowerup()
```
Dartemiss, line 117: This method can be const

Scripts/src/Script/PowerUp.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+void PowerUp::Spawn(const float3& spawn_position)
+{
+	EnemyDrop::Spawn(spawn_position);
+	SetupPowerupVfx();
+
+	picked = false;
+}
+
+void PowerUp::SetupPowerupVfx()
```
Dartemiss, line 143: Can be cosnt aswell

Dartemiss (changes requested):
GJ man!
I tested the branch and some things:
- The branch is very deprecated you have to merge develop here
- The level1/level2/boss scenes crashes because you just used level1_powerpus.scene for testing, please delete this scene from production
- I think the drops should have a lot more range to get collected like the current lightorbs otherwise is very tedious to get all of them (attracting_radio  should be bigger)
- The damage bonus is good but the invulnerable bonus doesn't have any feedback (or at least i couldn't see anything so I looked like a bug not receiving damage)
- The light ball doesn't have bonus damage?

Scripts/src/Script/PowerUp.cpp
JordiRomagosa: Yes, it is. In fact is very hard due to the drop rate, but for now I made it like this. To change it to "refresh duration" is as simple as changing the += for a =.

JordiRomagosa:
Realized requested changes and updated to develop.

Dartemiss (approved):
Okay looks good now!
Prepare some prefabs of the changes in the scenes. After we merge the big branches with the new level1/level2 and boss we'll merge this branch first after those.

As general comments I think that the red powerup is too long but is just my opinion and can be discussed for next delivery. 
GJ

PR #177, opened 2020-10-11
Knockup + LightShield rework + BossCurrentAttacks Tweaks

Dartemiss:
What's new?
- Knockup implemented on slam and phase2 attacks with @pikipoka animation.
- Lightball shield rebounds rework, now rebound the ball doesn't waste durability but now it goes faster and you have less time to press the button so it gets harder with each rebound.
- Some boss tweaks:
 Now on slam attack on multiplayer boss targets one of both players randomly.
 Now on phase2 guided crystals now only follows you until there are no more available crystals or all the players have been already hit, for balancing this damage have been increased.
 Now on avalanche attack if a rock above players is possible it will spawn there instead of going random.
 CrowdControl enum on boss_controller for chosing knockup or stun on attacks.
Little fix if player is knocked, doest accelerate.

PD: Sorry for this popurri PR my bad for not spliting features into different branches  :(


yalania (commented):
Remove the eeei.stm xD

Scripts/src/Script/PlayerMovement.h
```diff
@@ -152,6 +155,8 @@ class PlayerMovement : public Script
 	//Extra
 	float range_collision = 5.f;
 
+	mutable float next_pos_y = 0.f;
```
yalania, line 158: why this mutable?

yalania:
Some stuff I found:

- You can use the shield while you are doing the knockup animation
- Sometimes while doing the animation on a slam attack you can get a little bit stuck on the hand
- We need to remove the shield durability form ui

But really good job.

Scripts/src/Script/PlayerMovement.h
Dartemiss: because it was a debug variable that I used on a const method. I should remove this variable probably

Dartemiss:
I will delete the stm XD
I'll fix shield used on knockup
About getting stucked on the hand, I disable the colision of the hand while on knockup but then sometimes u fall there, i have to think on a better aproach
And the durability of shield is still used when u use the shield perse, I just did that the rebounds dont spend durability

yalania (approved):

Scripts/src/Script/BossAttackCrystalsGuided.cpp
```diff
@@ -92,10 +102,14 @@ void BossAttackCrystalsGuided::Update(float delta_time)
 	}
 
 	//If all projectiles have been thrown
-	if (current_number_of_projectiles_thrown >= number_of_projectiles_thrown)
+	if (current_number_of_projectiles_thrown >= number_of_projectiles_thrown | end)
```
mariofv, line 105: Bitwise intended?

Scripts/src/Script/BossAttackEarthquake.cpp
```diff
@@ -33,7 +33,7 @@ void BossAttackEarthquake::Update(float delta_time)
 	}
 	if (boss_controller->GetAnimationPercentage() > 0.70f) 
 	{
-		boss_controller->RotateBoss();
+		boss_controller->RotateBoss(0);
```
mariofv, line 36: Use the player enum :/ If it's unsigned, changed it to int

mariofv (approved):

PR #182, opened 2020-10-12
Add HitBlinker class

mariofv:
What's new:
- New HitBlinker, that triggers blink effect.
- Blink color and blink time can be customized.

Note: Put this script in a GameObject with a `ComponentMeshRenderer`, otherwise it won't work.


Dartemiss (changes requested):
GJ! There is a bug when player or enemy spawns while blinking. So a reset function would be needed

Dartemiss (approved):
Just adding scripts to levels is needed GJ

PR #183, opened 2020-10-14
Feature/audio fixes

RickyVimon:
What's new?
- SequenceController child on the AudioMannager to reset de sequences playing when respawning.
- Lvl2 ambient and crustal ambient sound (Corrupted/purified)

*Delete library before compiling


Dartemiss (approved):
AMAZING SPIDERMAN!

Scripts/src/Script/AudioManager.h
```diff
@@ -24,10 +24,13 @@ class AudioManager : public Script
 private:
 	void PlayAudioOnAwake();
 
+	void GetSequenceControler();
```
Pacasasgar, line 27: This function is not defined in the '.cpp'. Delete it if not being used.

RickyVimon: solved

Pacasasgar:
Appart from the comments I told you (solved already), very good job! Approving.

Pacasasgar (approved):

PR #184, opened 2020-10-15
Feature/static necroplant

RickyVimon:
What's new?

- Necroplants now have a checkbox Static which, if checked and "Alive" checkbox is also checked, the necroplant can be placed anywhere in the level and will not burry and teleport to other positions.
- Static necroplants will respawn again if the players also respawns (either on a checkpoint or at the beggining).
- 3 static Necroplants added during the Level 2 scene.


Scripts/src/Script/EnemyManager.cpp
```diff
@@ -359,4 +363,12 @@ bool EnemyManager::IsEnemyMeleeAttacking(EnemyController* enemy)
 bool EnemyManager::IsEnemyRangedAttacking(EnemyController* enemy)
 {
 	return static_cast<Mushdoom*>(enemy)->ranged_attack;
+}
+
+void EnemyManager::ResetStaticNecroplants()
+{
+	for (unsigned i = 0; i < static_necroplants.size(); ++i)
```
raulgonzalezupc, line 370: unsigned int

raulgonzalezupc (commented):
LGTM, but you have to change some things

Scripts/src/Script/EnemyManager.h
```diff
@@ -44,17 +45,21 @@ class EnemyManager : public Script
...
 	bool IsEnemyMeleeAttacking(EnemyController* enemy);
 	bool IsEnemyRangedAttacking(EnemyController* enemy);
 
 
+
 public:
 	std::vector<EnemyController*> enemies;
 	unsigned current_number_of_enemies_alive = 0;
 
+	std::vector<Necroplant*> static_necroplants;
```
raulgonzalezupc, line 61: I think you need to call this delete on the destructor

RickyVimon: Changes applied

Scripts/src/Script/EnemyManager.cpp
RickyVimon: done, thanks :)

raulgonzalezupc (approved):

Pacasasgar (approved):
LGTM, be careful with the scene conflicts when merging, take only one and then change that one.

PR #186, opened 2020-10-16
Fix/7 tutorial skippable fixed

raulgonzalezupc:
What is new?

- Tutorial Skipable UI
- Lightball given to the player when teleported
- Bug fix that the player could use the shield while on conversation


yalania (approved):
Nice job, just two minor bugs:

- If a tutorial popup is being show it will be place overlapping with the skip screen
- You don't need to make the player translate to the return position in case of skipping the tutorial.

Scripts/src/Script/AllyTrigger.cpp
```diff
@@ -188,3 +220,42 @@ void AllyTrigger::InitPublicGameObjects()
...
+		player_1->player_controller->SetIsInsideTrigger(false);
+		player_2->player_controller->SetIsInsideTrigger(false);
+	}
+	if (input_manager->GetGameInputDown("SkipTutorial", static_cast<PlayerID>(0)))
+	{
+		ally->do_tutorial = false;
+		ally->PlayAnimation(DialogueFeeling::FINISH);
+		tutorial_helper->tutorial_request->Disable();
+		ally->tutorial_request = false;
+		SetPlayersPositions(player_1_inside, player_2_inside);
```
yalania, line 255: You can remove the SetPlayersPositions from here I think.

raulgonzalezupc: I can do it for the don't do tutorial, but if you want to do it, does a random cut with the complain dialogue, so it's better to tp him. but i change it if don't

Scripts/src/Script/AllyInteractionsScript.cpp
```diff
@@ -318,6 +319,8 @@ void AllyInteractionsScript::TransformIntoLightBall()
 {
 	if (!ally_is_lightball)
 	{
+		App->engine_log->Log("throw light ball");
```
mariofv, line 322: Delete log before merging

Scripts/src/Script/AllyTrigger.cpp
```diff
@@ -188,3 +221,43 @@ void AllyTrigger::InitPublicGameObjects()
...
+		player_1->player_controller->SetIsInsideTrigger(false);
+		player_2->player_controller->SetIsInsideTrigger(false);
+	}
+	if (input_manager->GetGameInputDown("SkipTutorial", static_cast<PlayerID>(0)))
+	{
+		ally->do_tutorial = false;
+		ally->PlayAnimation(DialogueFeeling::FINISH);
+		tutorial_helper->tutorial_request->Disable();
+		ally->tutorial_request = false;
+		//SetPlayersPositions(player_1_inside, player_2_inside);
```
mariofv, line 257: Commented code should be deleted

mariofv (approved):
LGTM

PR #188, opened 2020-10-16
Recover of the feature/audio_volume

RickyVimon:
Use the Engine branch feature/audio_volume to test this one.
What's new?

New section on the Debug window with the volume control with a dragg int and two buttons to adjust the master volume of the game.


Pacasasgar:
As discussed on discord, I'm approving this. Nice job Ricky!

PD: Default volume in the debug menu appears to be '100', but in reality volume is at 50. This causes the volume to raise when moving from 100 to 99 in the slider. However that's not a big issue and volume in debug will be temporary, that's why I approve it anyway.

Pacasasgar (approved):

Scripts/src/Script/WorldManager.h
```diff
@@ -78,6 +78,9 @@ class WorldManager : public Script
 	//Tries for the game
 	int tries = 3;
 
+	//Audio Settings (To be changed once the UI Settings is done/merged
+	int master_volume = 50;
```
Dartemiss, line 82: unsigned maybe would be better

Dartemiss (approved):
LGTM

PR #192, opened 2020-10-18
fix lightball orbit around player

yalania:


mariofv (approved):
Works as intended!

JorxSS (approved):
LGTM

PR #197, opened 2020-10-19
Feature/biter feedback impovement

JordiRomagosa:
Impoved biter feedback though color changes and vfx effects.
Load the level1 biter scene to see the changes.

You need to use this PR in the engine to properly see the effects: https://github.com/OnionGalaxy/LittleOrionEngine/pull/290


Pacasasgar:
I would like to mention that it feels a bit strange when you hit the Biter and it turns white. When you hit a Mushdoom or the boss, it turns red just for half a second as feedback meaning you succeeded hitting. In this case, Biter turns white and the transition is too long also. Is this intended? If so, I would adjust this detail or discuss what can be done to improve this feeling when you successfully hit the biter.

However, I like the changes in general. It improves the Biter feedback for sure. Nice work!
As discussed in discord, use a prefab and put the Biter in all the scenes so we could do the final testing and approve it.

JordiRomagosa:
Already fixed the blink where it was causing problems, and also all scenes have the new biter.

PR #212, opened 2020-10-26
Fix/cleanup

yalania:
In general remove a ton of duplicate texture and old fbx that are not used anymore and are increasing the build size. I also removed some gameobejcts that should not be needed and adjust the code for it.

I tested it by doing a full round of the game from the menu - level1 to boss.


Dartemiss (approved):
LGTM, you have to merge with develop tho

raulgonzalezupc (approved):
works fine, just need the things from develop

PR #214, opened 2020-10-27
Fix/ powerup dops under or over terrain.

JordiRomagosa:
Fixed powerup spawning height.


Dartemiss (approved):
I did a full run and worked well. Only on boss sometimes the orbs went out of mesh/collider. And if you kill the enemies ouside the collider when they are comming to you also happens.

marcorod94 (approved):
It works great good Job

PR #215, opened 2020-10-27
Fix/starting point trails

JordiRomagosa:
Fixed trails for player dash and necroplant projectiles.


Misarte (approved):
Thank you Jordi for this fix! I appreciate! looks super good!

Dartemiss (approved):
LGTM

PR #218, opened 2020-10-30
animated statistics

vidalmelero:
what's new?

now statistics at the end of the level appear one by one.


Dartemiss (commented):
Looks nice but i would say that the time shouldnt keep counting, it is weird.

Dartemiss (approved):
LGTM

luismoyano (approved):
LGTM

PR #220, opened 2020-10-30
Feature/fix terrain damage

Dartemiss:
What's new?
- Refactor terrain damage computation, now bullet is not needed.
- New terrain damage type: Slowing water
- Level2 now is playable and easier as it uses slowing water mode instead of instakill water.

USE https://github.com/OnionGalaxy/LittleOrionEngine/pull/301


Scripts/src/Script/TerrainDamage.cpp
```diff
@@ -66,21 +71,35 @@ void TerrainDamage::Update()
...
+
+		if (player2_collides)
+		{
+			ComputeDamage(1);
+		}
+
+		if (terrain_type == TerrainType::SLOWING_WATER)
+		{
+			SlowPlayer(1, player2_collides, player2_in_range);
+		}
```
RickyVimon, line 102: wouldn't be easier to call this inside the if (playerx_collides) instead of using the playerx_collides bool?

RickyVimon (approved):
LGTM GJ Man!

Scripts/src/Script/TerrainDamage.cpp
```diff
@@ -199,10 +234,45 @@ void TerrainDamage::ComputeDamage(unsigned player)
...
+}
+
+bool TerrainDamage::IsPlayerInRange(GameObject* player) const
+{
+	float3 position = player->transform.GetGlobalTranslation();
+
+	return owner->aabb.global_bounding_box.maxPoint.x > position.x &&
+		owner->aabb.global_bounding_box.maxPoint.z > position.z &&
+		owner->aabb.global_bounding_box.minPoint.x < position.x &&
+		owner->aabb.global_bounding_box.minPoint.z < position.z;;
```
raulgonzalezupc, line 277: owner->aabb.global_bounding_box.minPoint.z < position.z;

Scripts/src/Script/TerrainDamage.cpp
Dartemiss: the problem is I need player2_collides and player2_in_range that are not on the ComputeDamage scope. But you are right it is weird so maybe I can make those variables members of the class.

Scripts/src/Script/TerrainDamage.cpp
Dartemiss: Done

raulgonzalezupc:
Works fine, but I personaly think that the slow on the water is too much. I would suggest to go a little bit quicker.

raulgonzalezupc (approved):

PR #223, opened 2020-11-01
Fix/boss cinematics

yalania:
- Added old normal for the water reflections ad light effect to be cooler.
- Fix boss's shadow being seen at the beginning of the cinematic.
- Fix uncorrupted effect on the outro cinematic.
- Tweak the outro cinematic for a cooler version where the lightball get bigger and remains in the boss position as the healed heart fo the forest.


Dartemiss (approved):
LGTM

tonifj (approved):
Good job! The cinematic looks very nice ^^

PR #224, opened 2020-11-03
Feature/new mushdoom attack

marcorod94:
Change the whole mushdoom scripts

- New battle circle and attack zones
- Configurable parameters from the enemy manager
- Move old Spawn zones
- Movement without bullet


JordiRomagosa (changes requested):
There are a few problems with the mushooms:

- They apear to be floating in the air:
image

- As you can see in this gif, they get stuck behind other mushdooms.
mushdooms

- Sometimes when moving or getting stuck, they do a "flickering". It looks as if they move slightly left-right with very short time.

- The seek behavior makes the mushdoom look like self-directed missiles... This is not a required change but it makes them very unrealistic. One suggestion for a fast improvement is to save the player's position each X time (for example 0.5 seconds) and use this saved position as the taget of the seek. This simple change should make them more realistic.
mushdooms2

RickyVimon (commented):
Looks Good to me but I see MushDooms getting another Mushdooms when chasing you

RickyVimon (approved):
LGTM Gj Man

JordiRomagosa:
I reviewed the changes and the previous problems are solved, but the seek continues to look awful.
Also, there is another major problem, and is that mushdooms traspass one another:
mushdooms2

marcorod94:
Well the overlapping is an action that is no going to be as common like on the bitters because they do overlap eachother

JordiRomagosa (approved):
Ok, if it is a design choice then I will approve it. Good job there remaking everything.

PR #225, opened 2020-11-04
Feature/shield update2.0

vidalmelero:
Whats new?

New shield bar style


Scripts/src/Script/LightShield.cpp
```diff
@@ -160,7 +171,7 @@ void LightShield::CastShield()
 void LightShield::UncastShield()
 {
 	active = false;
-
+	/*ui_manager->ShowShieldBar(active || broken, static_cast<uint32_t>(player_controller->player));*/
```
Dartemiss, line 174: if not used delete it

Dartemiss (approved):
GJ! I would say that I don't like that the bar is on your face when shield is destroyed recovering. I would prefer an icon on the hud that represents that without disturbing the gameplay. But it is just my opinion so I aprove!

luismoyano:
Aside from Enrique's comments, it looks good to me

luismoyano (approved):
LGTM

PR #227, opened 2020-11-05
Fix/38 stats total timer

Misarte:


Misarte:
The conflicts appear on how we display the time on the Text Field, someone coded simultaneously the field to look like: "00:00", while in my code is "0h 0min 0sec". Just let me know which one to keep, I don't mind.

Dartemiss:
I think that hours are not needed and regarding the format 00:00 or 0min 0sec,  I don't know pick whatever you prefer :D

Misarte:
Conflicts resolved guys! Please approve when you have time xD I kept the develop way of displaying time without the hours

Scripts/src/Script/WorldManager.cpp
```diff
@@ -84,12 +84,17 @@ void WorldManager::Awake()
 // Use this for initialization
 void WorldManager::Start()
 {
-
+	//StartTimer();
```
Dartemiss, line 87: if not used delete it

Scripts/src/Script/WorldManager.cpp
```diff
@@ -117,7 +122,7 @@ void WorldManager::Update()
 	if(loading && !App->resources->loading_thread_communication.loading)
 	{
 		on_pause = false;
-		StartTimer();
+		//StartTimer();
```
Dartemiss, line 125: if not use it delete it

Dartemiss (approved):

Scripts/src/Script/WorldManager.cpp
Misarte: Fixed!

yalania (approved):

PR #228, opened 2020-11-07
Skip Cinematics

marcorod94:
This is the new skip cinematics for the new cinematics system currently working on the boss scene 

you could test it by loading the boss scene from the main scene


JordiRomagosa (approved):
Looks good to me

Dartemiss (changes requested):
Works  fine excepts one thing. Sound. When you skip the cinematic, all boss sounds sound at the same time it is very weird. Also you can see how the boss is appears progressly instead of being there instantly.

JordiRomagosa:
> 
> 
> Works fine excepts one thing. Sound. When you skip the cinematic, all boss sounds sound at the same time it is very weird. Also you can see how the boss is appears progressly instead of being there instantly.

Oh, yeah, sorry... I tested without audio.

marcorod94:
The changes are done please review them

Dartemiss:
It stills sounds the boss screaming (is intended?) and also if I have the controller connected instead of the J button, a controller button should appear

marcorod94:
Fix the button error the total game pad wasn't been called, yes the sound is intended otherwise the scene will start with no music

Dartemiss:
ALso, if cinematic is not skipped the indicator remains.

Dartemiss (approved):
LGTM

PR #232, opened 2020-11-14
pause screen refactor

vidalmelero:
new art for pause screen:
cursor deleted, new images.


mariofv (changes requested):
Controls sub window has incorrect back button (it should be B instead of A) and icons doesn't change if you plug/unplug the controller. Also is not possible to swap to keyboard controlls.

The general pause frame also changes when you move up/down, it should be independent of input

vidalmelero:
> Controls sub window has incorrect back button (it should be B instead of A) and icons doesn't change if you plug/unplug the controller. Also is not possible to swap to keyboard controlls.
> 
> The general pause frame also changes when you move up/down, it should be independent of input

PR #235, opened 2020-11-15
Fix on waves not being triggered due to difficulty settings

luismoyano:
Tested and tuned in all difficulties


Dartemiss:
As this feature is not completed (no UI) I'll  close the PR

luismoyano:
@Dartemiss This is not a feature, it's a fix. The UI should be already in develop

Dartemiss:
It isn't

